PR #58, opened 2020-05-10
Use `desktop-app` package for desktop integration

chrisjbillington:
Instead of `labscript_utils.winshell` and `labscript_utils.winlauncher`.

OS menu (e.g. Start menu on Windows) shortcuts can be created with
`desktop-app install lyse`

And lyse can be run from the terminal with `lyse` or
`lyse-gui` to run without a console on Windows.


rpanderson:
See chrisjbillington/desktop-app#1.

PR #65, opened 2020-05-29
Fix #63 - matplotlib 3.2 compatibility

### LINKED ISSUE #63: Matplotlib 3.2.x Compatability Issue
When starting lyse (hg commit d8c69309a825) in an environment with matplotlib 3.2.1 I get the following error:

```python
Traceback (most recent call last):
  File "C:\Users\waveguide3\Desktop\labscript_test_install_3\lyse\analysis_subprocess.py", line 505, in <module>
    lyse.figure_manager.install()
  File "C:\Users\waveguide3\Desktop\labscript_test_install_3\lyse\figure_manager.py", line 139, in install
    raise RuntimeError(message)
RuntimeError: install() must be imported prior to importing pylab/pyplot in order to correctly override the figure() function.
```

In the latest github commit those lines are here https://github.com/labscript-suite/lyse/blob/622b79eb4bff4f7cc784447ce73da78b4d094297/lyse/analysis_subprocess.py#L444 and here https://github.com/labscript-suite/lyse/blob/622b79eb4bff4f7cc784447ce73da78b4d094297/lyse/figure_manager.py#L133

This error doesn't occur with matplotlib 3.1.3. I opened up the debugger and it seems that with matplotlib 3.2.1 `'matplotlib.pyplot' in sys.modules` evaluates to `False` before the call to `import matplotlib` in `analysis_subprocess.py` but it evaluates to `True` right after the import.

One work around may be to put the `import lyse...` block of code before the `import matplotlib...` block of code in `analysis_subprocess.py`. That worked fine with both matplotlib 3.1.3 and 3.2.1, though I haven't tested older versions.

Interestingly lyse also seemed to work just fine with matplotlib 3.2.1 (without the above change) if I simply commented out the line that raises the `RuntimeError`. Admittedly I'm not 100% sure what behavior to look for to ensure that the `FigureManager` monkey patch works. The figure windows aren't brought to the front when new data is plotted if that's relevant.

Sticking with matplotlib 3.1.3 for now is no problem at all; I'm sure you guys have plenty on your plates at the moment with the bitbucket move.

chrisjbillington:
Set matplotlib backend via an environment variable so that we don't need
to import matplotlib prior to `figure_manager.install()`

Drop obsolete PyQt4 compat.

### DIFF
--- a/lyse/analysis_subprocess.py
+++ b/lyse/analysis_subprocess.py
@@ -21,7 +21,7 @@ import traceback
 import time
 from types import ModuleType
 
-from qtutils.qt import QtCore, QtGui, QtWidgets, QT_ENV, PYQT5
+from qtutils.qt import QtCore, QtGui, QtWidgets
 from qtutils.qt.QtCore import pyqtSignal as Signal
 from qtutils.qt.QtCore import pyqtSlot as Slot
 
@@ -431,11 +431,7 @@ class AnalysisWorker(object):
         
 if __name__ == '__main__':
 
-    import matplotlib
-    if QT_ENV == PYQT5:
-        matplotlib.use("QT5Agg")
-    else:
-        matplotlib.use("QT4Agg")
+    os.environ['MPLBACKEND'] = "qt5agg"
 
     import lyse
     from lyse import LYSE_DIR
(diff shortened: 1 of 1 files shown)


philipstarkey:
Is this expected to be backwards compatible with older matplotlib or should we add a matplotlib>=3 requirement as part of this PR?

chrisjbillington:
The environment variable has existed [since](https://github.com/matplotlib/matplotlib/commit/0ead4b2862fc91613bc8f17e1979ec750f076964) v1.5 5 years ago so I think we're good.

PR #71, opened 2020-06-24
Run.get_* methods should open the HDF5 file in read-only mode

philipstarkey:
Fixes #42

Also addressed labscript-suite/labscript-utils#47 (removing h5py deprecation warnings) for this module

Also fixed an issue where calling `set_group` on a `Run` inside a `Sequence` could disable the read-only mode of the `Run` object.

Added `no_write` support to `Sequence`.

PR #72, opened 2020-07-16
Make Run.set_group() create the group if necessary.

zakv:
Calling `Run.set_group()` used to create the group if it didn't already exist:

https://github.com/labscript-suite/lyse/blob/74dfc19a8b3bcfeddaff63d11f21ca847184a36b/lyse/__init__.py#L133-L137

However after https://github.com/labscript-suite/lyse/pull/71, `Run.set_group()` only makes the `'/results'` group:

https://github.com/labscript-suite/lyse/blob/f7a14277e45f57c34f71ccd7d85fb072c67ce412/lyse/__init__.py#L148-L150

This PR just changes the last line in that excerpt to `self._create_group_if_not_exists(self.h5_path, '/results', self.group)` to restore the previous behavior.

### DIFF
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -147,7 +147,7 @@ class Run(object):
 
     def set_group(self, groupname):
         self.group = groupname
-        self._create_group_if_not_exists(self.h5_path, '/', 'results')
+        self._create_group_if_not_exists(self.h5_path, '/results', self.group)
         # restore no_write attribute now we have set the group
         if self._no_group is not None and self._no_group[0]:
             self.no_write = self._no_group[1]


philipstarkey:
Looks like a copy-paste bug when doing #71 (I suspect I copied the line from earlier in the code and forgot to update which group it was creating). 

@chrisjbillington Guessing this should be cherry picked over to the maintenance branch and released as v3.0.1?

chrisjbillington:
Yep! If we had more bug fixes in the pipeline we would want to make a bugfix release for more than just one bug, but there isn't anything else so might as well release with just this bugfix

philipstarkey:
So actually, in reviewing this again before cherry picking I realised there are further issues now (see #78).

@zakv Do you feel like looking into this at some point?

zakv:
Will do!

PR #73, opened 2020-10-09
Bugfix for sequence

lakodarmstadt:
When instantiating a sequence, usually a _result.h5-file was created to save sequence results. However, currently h5-files are only opened in read mode via h5py.File(filename, 'r') which is not creating new files if they do not exist. This is solved by first trying to open the cooresponding h5-file in read mode and is an exeption is thrown to open it in write mode. This is only added in the sequence instatiation as single shots have there shot-h5-file.

### DIFF
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -246,6 +246,7 @@ class Run(object):
             if not group:
                 # Save dataset to results group by default
                 group = 'results/' + self.group
+                
             elif not group in h5_file:
                 # Create the group if it doesn't exist
                 h5_file.create_group(group) 
@@ -422,6 +423,11 @@ class Sequence(Run):
             run_paths = run_paths['filepath']
         self.h5_path = h5_path
         self.no_write = no_write
+        try:
+            h5py.File(h5_path, 'r')
+        except:
+            h5py.File(h5_path,'w')
+
         if not self.no_write:
             self._create_group_if_not_exists(h5_path, '/', 'results')
                  


lyse/__init__.py
```diff
@@ -246,6 +246,7 @@ def save_result_array(self, name, data, group=None,
             if not group:
                 # Save dataset to results group by default
                 group = 'results/' + self.group
+
```
philipstarkey, line 249: Could you remove this unnecessary change so that it doesn't clutter up the diff?

lyse/__init__.py
```diff
@@ -422,6 +423,11 @@ def __init__(self, h5_path, run_paths, no_write=False):
             run_paths = run_paths['filepath']
         self.h5_path = h5_path
         self.no_write = no_write
+        try:
+            h5py.File(h5_path, 'r')
```
philipstarkey, line 427: I think that this may leave an open reference to the h5 file. It would be better to use a context manager like this: 

``` 
with h5py.File(h5_path, 'r') as f:
    pass
```
Which ensures that the file is closed appropriately.

lyse/__init__.py
```diff
@@ -422,6 +423,11 @@ def __init__(self, h5_path, run_paths, no_write=False):
...
+        except:
```
philipstarkey, line 428: It is best to catch a specific exception class here. Could you catch the specific exception raised? (Presumably it is `FileNotFoundError`?)

lyse/__init__.py
```diff
@@ -422,6 +423,11 @@ def __init__(self, h5_path, run_paths, no_write=False):
...
+            h5py.File(h5_path,'w')
```
philipstarkey, line 429: Similar to previous comment, use of a context manager here would be preferable. Also, you might want to consider using the `'a'` filemode (so that the file doesn't get truncated should it happen to be created by an external process in between the two h5py calls executing - which could conceivably be a significant time should there be contention on the zlock/h5lock lock).

I think it might also be appropriate to wrap the creation call in a check of the `no_write` argument and raise the `FileNotFoundError` should the file not exist and the `no_write` argument be true?

philipstarkey (changes requested):
Thanks for working on this! I've left a few comments about potential improvements to the code to make it more robust against edge-cases.

philipstarkey:
Due to inactivity, I'm closing this in favour of #80 which also includes this fix.

PR #74, opened 2020-10-21
Add n_runs column to lyse dataframe

zakv:
In my lab we often want to find out if all the shots from a click on "Engage" in runmanager have finished. To do this we need to compare the `n_runs` and `run number` attributes of the root directory of a shot's hdf5 file.

It would be convenient if both of those pieces of information were included in the lyse dataframe, and that could make it easier for new users to find that info as well. The lyse dataframe already has a `run number` column, so this PR simply adds an `n_runs` column to the lyse dataframe.


philipstarkey:
Since this is pretty trivial, I'm merging.

PR #76, opened 2020-10-29
Dataframe subset

zakv:
This PR implements ~some of~ the features requested in #3 by adding optional `n_sequences` and `filter_kwargs` arguments to `lyse.data()`.

Changes:

* Add an optional `n_sequences` argument to `lyse.data()` which allows the user to request only a subset of the lyse dataframe, which can speed execution.
  * If there are fewer than `n_sequences` sequences available, then the entire lyse dataframe is returned.
* Add an optional `filter_kwargs` argument to `lyse.data()`
  * This is a dictionary of keyword arguments passed to `Dataframe.filter()` before the dataframe is transmitted.
* Improve thread safety of `Webserver.handler()`
* Add a docstring for `lyse.data()`

### DIFF
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -103,15 +103,19 @@ def data(filepath=None, host='localhost', port=_lyse_port, timeout=5, n_sequence
             42519 if the labconfig has no such entry.
         timeout (float, optional): The timeout, in seconds, for the
             communication with lyse. Defaults to 5.
-        n_sequences (int, optional): The number of sequences to include in the
-            returned dataframe where one sequence corresponds to one call to
-            engage in runmanager. The dataframe rows for the most recent
-            `n_sequences` sequences are returned. If set to `None`, then all
-            rows are returned. Defaults to `None`.
+        n_sequences (int, optional): The maximum number of sequences to include
+            in the returned dataframe where one sequence corresponds to one call
+            to engage in runmanager. The dataframe rows for the most recent
+            `n_sequences` sequences are returned. If the dataframe contains
+            fewer than `n_sequences` sequences, then all rows will be returned.
+            If set to `None`, then all rows are returned. Defaults to `None`.
 
     Raises:
         ValueError: If `n_sequences` isn't `None` or a nonnegative integer, then
-            a `ValueError` is raised.
+            a `ValueError` is raised. Note that no `ValueError` is raised if
+            `n_sequences` is greater than the number of sequences available. In
+            that case as all available sequences are returned, i.e. the entire
+            lyse dataframe is returned.
 
     Returns:
         (pandas series or dataframe): If `filepath` is provided, then a pandas
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -19,6 +19,7 @@ import pickle as pickle
 import inspect
 import sys
 import threading
+import shlex
 
 import labscript_utils.h5_lock, h5py
 from labscript_utils.labconfig import LabConfig
@@ -76,7 +77,7 @@ class _RoutineStorage(object):
 routine_storage = _RoutineStorage()
 
 
-def data(filepath=None, host='localhost', port=_lyse_port, timeout=5, n_sequences=None):
+def data(filepath=None, host='localhost', port=_lyse_port, timeout=5, n_sequences=None, filter_kwargs=None):
     """Get data from the lyse dataframe or a file.
     
     This function allows for either extracting information from a run's hdf5
@@ -126,14 +127,22 @@ def data(filepath=None, host='localhost', port=_lyse_port, timeout=5, n_sequence
     if filepath is not None:
         return _get_singleshot(filepath)
     else:
-        command = 'get dataframe'
+        command_list = ['get_dataframe']
         if n_sequences is not None:
             if type(n_sequences) is int and n_sequences >= 0:
-                command = command + ' n_sequences={n_sequences}'.format(n_sequences=n_sequences)
+                command_list.append('--n_sequences {n_sequences}'.format(n_sequences=n_sequences))
             else:
                 msg = """n_sequences must be None or an integer greater than 0 but 
                     was {n_sequences}.""".format(n_sequences=n_sequences)
                 raise ValueError(dedent(msg))
+        if filter_kwargs is not None:
+            if type(filter_kwargs) is dict:
+                command_list.append('--filter_kwargs ' + shlex.quote(repr(filter_kwargs)))
+            else:
+                msg = """filter must be None or a dictionary but was 
+                    {filter_kwargs}.""".format(filter_kwargs=filter_kwargs)
+                raise ValueError(dedent(msg))
+        command = ' '.join(command_list)
         df = zmq_get(port, host, command, timeout)
 
         try:
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -125,8 +125,8 @@ def data(filepath=None, host='localhost', port=_lyse_port, timeout=5, n_sequence
             returned. For example to call `filter()` with `like='temperature'`,
             set `filter_kwargs` to `{'like':'temperature'}`. If set to `None`
             then `Dataframe.filter()` will not be called. See
-            `Dataframe.filter()`'s documentation for more information. Defaults
-            to `None`.
+            :meth:`pandas:pandas.DataFrame.filter` for more information.
+            Defaults to `None`.
 
     Raises:
         ValueError: If `n_sequences` isn't `None` or a nonnegative integer, then
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -81,8 +81,8 @@ def data(filepath=None, host='localhost', port=_lyse_port, timeout=5, n_sequence
         return _get_singleshot(filepath)
     else:
         command = 'get dataframe'
-        if n_sequences:
-            if type(n_sequences) is int and n_sequences > 0:
+        if n_sequences is not None:
+            if type(n_sequences) is int and n_sequences >= 0:
                 command = command + ' n_sequences={n_sequences}'.format(n_sequences=n_sequences)
             else:
                 msg = """n_sequences must be None or an integer greater than 0 but 
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -162,24 +162,37 @@ def data(filepath=None, host='localhost', port=_lyse_port, timeout=5, n_sequence
                 raise ValueError(dedent(msg))
         command = ' '.join(command_list)
         df = zmq_get(port, host, command, timeout)
-
-        try:
-            padding = ('',)*(df.columns.nlevels - 1)
-            try:
-                integer_indexing = _labconfig.getboolean('lyse', 'integer_indexing')
-            except (LabConfig.NoOptionError, LabConfig.NoSectionError):
-                integer_indexing = False
-            if integer_indexing:
-                df.set_index(['sequence_index', 'run number', 'run repeat'], inplace=True, drop=False)
-            else:
-                df.set_index([('sequence',) + padding,('run time',) + padding], inplace=True, drop=False)
-                df.index.names = ['sequence', 'run time']
-        except KeyError:
-            # Empty DataFrame or index column not found, so fall back to RangeIndex instead
-            pass
+        # Ensure conversion to multiindex is done, which needs to be done here
+        # if the server is running an old version of lyse.
+        _rangeindex_to_multiindex(df, inplace=True)
         df.sort_index(inplace=True)
         return df
-        
+
+def _rangeindex_to_multiindex(df, inplace):
+    if isinstance(df.index, pandas.MultiIndex):
+        # The dataframe has already been converted.
+        return df
+    try:
+        padding = ('',)*(df.columns.nlevels - 1)
+        try:
+            integer_indexing = _labconfig.getboolean('lyse', 'integer_indexing')
+        except (LabConfig.NoOptionError, LabConfig.NoSectionError):
+            integer_indexing = False
+        if integer_indexing:
+            out = df.set_index(['sequence_index', 'run number', 'run repeat'], inplace=inplace, drop=False)
+            # out is None if inplace is True, and is the new dataframe is inplace is False.
+            if not inplace:
+                df = out
+        else:
+            out = df.set_index([('sequence',) + padding,('run time',) + padding], inplace=inplace, drop=False)
+            if not inplace:
+                df = out
+            df.index.names = ['sequence', 'run time']
+    except KeyError:
+        # Empty DataFrame or index column not found, so fall back to RangeIndex instead
+        pass
+    return df
+
 def globals_diff(run1, run2, group=None):
     return dict_diff(run1.get_globals(group), run2.get_globals(group))
  
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -76,11 +76,20 @@ class _RoutineStorage(object):
 routine_storage = _RoutineStorage()
 
 
-def data(filepath=None, host='localhost', port=_lyse_port, timeout=5):
+def data(filepath=None, host='localhost', port=_lyse_port, timeout=5, n_sequences=None):
     if filepath is not None:
         return _get_singleshot(filepath)
     else:
-        df = zmq_get(port, host, 'get dataframe', timeout)
+        command = 'get dataframe'
+        if n_sequences:
+            if type(n_sequences) is int and n_sequences > 0:
+                command = command + ' n_sequences={n_sequences}'.format(n_sequences=n_sequences)
+            else:
+                msg = """n_sequences must be None or an integer greater than 0 but 
+                    was {n_sequences}.""".format(n_sequences=n_sequences)
+                raise ValueError(dedent(msg))
+        df = zmq_get(port, host, command, timeout)
+
         try:
             padding = ('',)*(df.columns.nlevels - 1)
             try:
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -77,6 +77,48 @@ routine_storage = _RoutineStorage()
 
 
 def data(filepath=None, host='localhost', port=_lyse_port, timeout=5, n_sequences=None):
+    """Get data from the lyse dataframe or a file.
+    
+    This function allows for either extracting information from a run's hdf5
+    file, or retrieving data from lyse's dataframe. If `filepath` is provided
+    then data will be read from that file and returned as a pandas series. If
+    `filepath` is not provided then the dataframe in lyse, or a portion of it,
+    will be returned.
+
+    Args:
+        filepath (str, optional): The path to a run's hdf5 file. If a value
+            other than `None` is provided, then this function will return a
+            pandas series containing data associated with the run. In particular
+            it will contain the globals, singleshot results, multishot results,
+            etc. that would appear in the run's row in the Lyse dataframe, but
+            the values will be read from the file rather than extracted from the
+            lyse dataframe. If `filepath` is `None, then this function will
+            instead return a section of the lyse dataframe. Note that if
+            `filepath` is not None, then the other arguments will be ignored.
+            Defaults to `None`.
+        host (str, optional): The address of the computer running lyse. Defaults
+            to `'localhost'`.
+        port (int, optional): The port on which lyse is listening. Defaults to
+            the entry for lyse's port in the labconfig, with a fallback value of
+            42519 if the labconfig has no such entry.
+        timeout (float, optional): The timeout, in seconds, for the
+            communication with lyse. Defaults to 5.
+        n_sequences (int, optional): The number of sequences to include in the
+            returned dataframe where one sequence corresponds to one call to
+            engage in runmanager. The dataframe rows for the most recent
+            `n_sequences` sequences are returned. If set to `None`, then all
+            rows are returned. Defaults to `None`.
+
+    Raises:
+        ValueError: If `n_sequences` isn't `None` or a nonnegative integer, then
+            a `ValueError` is raised.
+
+    Returns:
+        (pandas series or dataframe): If `filepath` is provided, then a pandas
+            series with the data read from that file is returned. If `filepath`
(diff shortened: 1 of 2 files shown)


zakv:
I actually have a few questions still:

1. Previously `WebServer.handler()` didn't have anything special in it to make it thread safe. Is it necessary to add a lock or an `inmain_decorator` or something now that there is some processing of the dataframe's info in `WebServer.handler()`?
1. I'd be happy to implement the other feature requested in #3 as well, namely an optional `date` argument. I just have two questions related to that:
    * What form should the argument take? A string with some particular formatting? An instance of some date/time class?
    * Should this be AND'ed with `n_sequences`? In other words if `n_sequences = 100` and `date='today'` then only return sequences from today, up to 100 of them.

philipstarkey:
Yeah, I believe I've raised the thread safety issue somewhere before (doesn't seem to have its own issue though...). I'm pretty sure the existing implementation is thread unsafe because the data frame is sent to internal zprocess methods for serialisation (I think using pickle?), and it's passed by reference. So even a lock or in_main call won't help with that because the lock would always be released before zprocess serialised it.

The only way to safe with the current architecture is to do a deepcopy of the data frame and return that (assuming it's even possible to deepcopy a data frame). Performance would probably be terrible for large data frames? Maybe it's worth trying though? I also have no idea if subsets return deep copies or only shallow copies of data.

Regarding 2, what about having n_sequences and then a set of arguments to pass directly to the filter method of the data frame? Then we don't have to worry about what sort of filtering we're supporting. Maybe have a single argument for `data()` called `filter` which the user would specify as a dictionary where the key/values match the pandas method and then the lyse code can just `**filter_kwargs` inside the call to `filter` if that makes sense?

chrisjbillington:
zprocess doesn't defer serialisation such that it might occur in a different thread - serialisation occurs in the calling thread. (zmq actually sending the data is deferred, but not serialisation via pickling). So a lock or inmain() should be sufficient I should think.

pickling holds the GIL, so the dataframe can't be modified mid-pickle. However, the dataframe that gets sent may be in the middle of being updated by `update_row()`, so it may have some analysis results updated and some not yet.

I think `inmain_decotrator` on the handler would remove that possibility, since `update_row()` called in the main thread.

philipstarkey:
Adding `inmain_decorator` on the handler only prevents updates to the dataframe while that method runs though. All of the machinery in `ZMQServer.mainloop` and ZMQ itself (where the pickling occurs right?) runs after `ZMQServer.handler` completes, and so the main thread can continue and may (partially) modify the dataframe before it is sent down the socket (unless we make a copy). So I don't think that solves the thread safety problem.

Also, just because pickling holds the GIL, doesn't Python still context switch between threads every 15ms anyway (previously was 1000 lines of bytecode in Python 2)? I don't think relying on the GIL is sufficient.

Both issues can be solved with a dedicated DataFrame lock, but zprocess would need to be updated so it can also be passed the lock somehow (to be held while internal zprocess/zmq code interacts with the dataframe). To be honest, I'm surprised there isn't a PEP for automatically using an `RLock` before accessing an object if one exists in `object.__lock__`. That way code like zprocess wouldn't need to grow additional kwargs to pass around lock objects...

chrisjbillington:
Ah I'm sorry, you're right. Would have to have the zmqserver's mainloop hold a lock somehow, not currently supported or easy to do by subclassing.

No, no other Python code can run whilst pickling, since pickling is a single python bytecode instruction that calls into C, and doesn't release the GIL within the C code. The whole interpreter stops whilst pickling, it's not possible for an object to be modified mid-pickle. I think this is by design and not just an implementation detail, though I'm not 100% sure.

But that's not super relevant - pickling itself can't fail, but it can begin at any arbitrary moment whilst the dataframe is being modified, so you can still get a half-modified dataframe.

> EDIT: alternatively, if handler returns a copy of the dataframe, made while run in the main thread, then that would be safe too (since the only reference to the copy held after handler runs would be in the internal zprocess code). I was hoping to avoid data duplication though.

To avoid data duplication, we could add  `inmain_decorator` to the handler, and have it call the ZMQServer's send method directly to send the response then and there instead of returning the object to be sent. Then we instead return ZMQServer.NO_RESPONSE:

```python
@inmain_decorator(wait_for_return=True)
def handler(self, data):
    ....
    self.send(dataframe)
    return self.NO_RESPONSE
    ...
```

I use this same mechanism in the IMQDdxCamera BLACS tab, there it was because the handler does some time-consuming work updating the displayed image, but I didn't want that to delay the response. So it calls `send()` to send a response immediately before displaying the new image.

zakv:
I've implemented Chris's suggestion with `@inmain_decorator` and `self.send()`.

I also implemented Phil's `filter` suggestion, though I called it `filter_kwargs` to avoid using the name of a builtin. Sending a dictionary between the processes was a little tricky so I had to bring in a few imports: `argparse`, `shlex`, and `ast`. They're all from the standard library though.

I noticed that the other classes were using `ParentClassName.__init__()` rather than `super().__init__()` so I did the same with `WebServer`.

There is one issue that I haven't figured out yet though. For some reason the dataframe's index is replaced with a `RangeIndex` if `filter_kwargs` is used:

```python
In []: filter_kwargs={'like':'run'}

In []: lyse.data(filter_kwargs=filter_kwargs).index
Out[]: RangeIndex(start=0, stop=3, step=1)

In []: lyse.data().filter(**filter_kwargs).index
Out[]: 
MultiIndex([('2020-10-29 19:07:50-04:00', '2020-10-29 19:07:51-04:00'),
            ('2020-10-29 19:07:50-04:00', '2020-10-29 19:07:53-04:00'),
            ('2020-10-29 19:07:51-04:00', '2020-10-29 19:07:54-04:00')],
           names=['sequence', 'run time'])
```
At first I assumed it had to do with the way that `filter_kwargs` is sent over ZeroMQ to the `WebServer`, but I saw the same behavior even when I temporarily hardcoded a value for `filter_kwargs` in `WebServer.handler()`. I'll keep digging but any suggestions would be appreciated.

zakv:
Ok I think I see the issue actually. The `app.filebox.shots_model.dataframe` uses a `RangeIndex`, then when the dataframe reaches `lyse.data()` its changed to a multiindex. Running `filter()` in `WebServer.handler` runs it on `app.filebox.shots_model.dataframe` and can end up filtering out the columns that would later be used to generate the multiindex. Then `lyse.data()` swallows the error generated when trying to make the multiindex and just returns the dataframe with the rangeindex.

I guess the solution is to move the multiindex-generating code from `lyse.data()` to `WebServer.handler()`? That way it could be run before `filter()`.

philipstarkey:
Was the use of shlex etc. to maintain backwards/forwards compatibility when accessing the dataframe from a standalone script? (For example lyse v3.0 server talking to lyse v3.1 client on another PC?). 

Normally I would have probably just sent something like `['get_dataframe', n_sequences, filter_kwargs]` as the `command` passed to `zmq_get` (zmq_get can send python objects), but I guess that would mean new lyse clients would raise an error when talking to old lyse servers. Oh...if n_sequences is None and filter_kwargs is None, the new client could fall back to sending just `'get_dataframe'` instead of a list? The other direction could be accommodated too (old lyse clients talking to new lyse servers) by using a more complex if statement on the server side (to capture both `'get_dataframe'` and `['get_dataframe', ..]`. So you get an (unhelpful) error message if you use new client with old server and take advantage of the new kwargs (although could we catch the error returned by the old ZMQ server and present a more useful error to the user?)

Thoughts? 

Regarding the `RangeIndex`, there is a comment about falling back to `RangeIndex` just below the call to `zmq_get` in `data()`. I'm not familiar with the piece of code, but just noticed it when looking at the other stuff now. Might be worth seeing if the try/except is suppressing some useful debug information there.

philipstarkey:
> Ok I think I see the issue actually. The `app.filebox.shots_model.dataframe` uses a `RangeIndex`, then when the dataframe reaches `lyse.data()` its changed to a multiindex. Running `filter()` in `WebServer.handler` runs it on `app.filebox.shots_model.dataframe` and can end up filtering out the columns that would later be used to generate the multiindex. Then `lyse.data()` swallows the error generated when trying to make the multiindex and just returns the dataframe with the rangeindex.
> 
> I guess the solution is to move the multiindex-generating code from `lyse.data()` to `WebServer.handler()`? That way it could be run before `filter()`.

Missed this while writing my own reply. Yeah, that could be a good idea. It does seem like filter should be run after generating the multiindex. Might need to test how this interacts with v3.0 server taking to v3.1 client though (and 3.1 server talking to 3.0 client). The code to convert to multi-index may need to be in both client and server until 4.0?

zakv:
> Was the use of shlex etc. to maintain backwards/forwards compatibility when accessing the dataframe from a standalone script? (For example lyse v3.0 server talking to lyse v3.1 client on another PC?).

Nope, somehow the idea that I could send anything besides a string didn't occur to me. That's a much better idea so I'll switch it over to do that instead and get rid of the superfluous imports.

> Normally I would have probably just sent something like `['get_dataframe', n_sequences, filter_kwargs]` as the `command` passed to `zmq_get` (zmq_get can send python objects), but I guess that would mean new lyse clients would raise an error when talking to old lyse servers. Oh...if n_sequences is None and filter_kwargs is None, the new client could fall back to sending just `'get_dataframe'` instead of a list? The other direction could be accommodated too (old lyse clients talking to new lyse servers) by using a more complex if statement on the server side (to capture both `'get_dataframe'` and `['get_dataframe', ..]`. So you get an (unhelpful) error message if you use new client with old server and take advantage of the new kwargs (although could we catch the error returned by the old ZMQ server and present a more useful error to the user?)

Yeah that is a tricky situation which I hadn't considered. I can't think of a better solution than what you've suggested here. Actually the old lyse server just ignores the request if `command` is a list, so `zmq_get()` returns `None` in that case. That makes it easy to notice and raise a helpful error client-side if the server is outdated and doesn't support the new arguments. It can also fall back to sending just `'get dataframe'` if `n_sequences` and `filter_kwargs` are `None`.

> It does seem like filter should be run after generating the multiindex. Might need to test how this interacts with v3.0 server taking to v3.1 client though (and 3.1 server talking to 3.0 client). The code to convert to multi-index may need to be in both client and server until 4.0?

Yep I think you're right.

lyse/__init__.py
```diff
@@ -76,28 +76,132 @@ class _RoutineStorage(object):
...
+            to engage in runmanager. The dataframe rows for the most recent
+            `n_sequences` sequences are returned. If the dataframe contains
+            fewer than `n_sequences` sequences, then all rows will be returned.
+            If set to `None`, then all rows are returned. Defaults to `None`.
+        filter_kwargs (dict, optional): A dictionary of keyword arguments to
+            pass to the `Dataframe.filter()` method before the lyse dataframe is
+            returned. For example to call `filter()` with `like='temperature'`,
+            set `filter_kwargs` to `{'like':'temperature'}`. If set to `None`
+            then `Dataframe.filter()` will not be called. See
+            `Dataframe.filter()`'s documentation for more information. Defaults
```
philipstarkey, line 128: You should be able to generate a link to the pandas documentation using an intersphinx mapping. It's already configured in the sphinx conf.py, you just need to work out what magic string to put in here to make it work! [This](https://sphobjinv.readthedocs.io/en/latest/) may be of use!

lyse/__init__.py
```diff
@@ -76,28 +76,132 @@ class _RoutineStorage(object):
...
+            command = 'get dataframe'
+        else:
+            command = ('get dataframe', n_sequences, filter_kwargs)
+        df = zmq_get(port, host, command, timeout)
+        if isinstance(df, str) and df.startswith('error: operation not supported'):
+            # Sending a tuple for command to an outdated lyse servers causes it
+            # to reply with an error message.
+            msg = """The lyse server does not support n_sequences or filter_kwargs.
+                Call this function without providing those arguments to communicate
+                with this server."""
```
philipstarkey, line 172: Consider adding an additional sentence suggesting an alternative to removing the unsupported options is to upgrade the version of the lyse server?

lyse/__main__.py
```diff
@@ -152,18 +152,43 @@ def get_screen_geometry():
...
+            _, n_sequences, filter_kwargs = request_data
             # infer_objects() picks fixed datatypes for columns that are compatible with
             # fixed datatypes, dramatically speeding up pickling. It is called here
             # rather than when updating the dataframe as calling it during updating may
             # call it needlessly often, whereas it only needs to be called prior to
             # sending the dataframe to a client requesting it, as we're doing now.
             app.filebox.shots_model.infer_objects()
-            return app.filebox.shots_model.dataframe
+            df = app.filebox.shots_model.dataframe
+            df = _rangeindex_to_multiindex(df, inplace=False)
```
philipstarkey, line 169: So the use of `inplace=False` here makes perfect sense, but does that mean that we have a deepcopy of the dataframe? And thus actually only need to lock this one line with `inmain`, since we're copying anyway?

Or should we be trying to avoid that performance hit and only doing the index conversion server side if `filter_args` is not None? That means `_rangeindex_to_multiindex` would need to detect if it has already run once so it could always be called on the client side.

I'm not quite sure what the best approach here is.

lyse/__init__.py
philipstarkey: Oh, there's no sphinx docs configured to show this anyway...not worth adding that into this same PR. I think you can just replace `Dataframe.filter()` with ``:meth:`pandas:pandas.DataFrame.filter` `` and be done with it.

lyse/__init__.py
zakv: Good idea, this message is updated in 4c944dd.

lyse/__main__.py
zakv: I've spent a few hours trying to figure out pandas' memory management but haven't been able to confidently come to a conclusion about this. The topic of "when does panda return a copy vs return a view" is a bit complicated.

Playing around in the interpreter, I tried the following:

```python
df0 = lyse.data()
df0 = df1.set_index('sequence', inplace=False, drop=False)
df1.loc[:, 'run number'] = 12345
```

That last line modifies `df1` but `df0` remains unchanged, so it seems like `set_index()` did make a deep copy. I suspect that it is guaranteed to always make a deep copy, but I'm definitely not 100% sure. I'm also not sure if it is an immediate copy, or a copy-on-write. If it's a copy-on-write then in `WebServer.handler()` the underlying data (ideally) shouldn't be duplicated since the dataframe copy goes out of scope when `handler()` returns, which is before anything edits the underlying data thanks to the `@inmain_decorator`.

We could make lyse only do the conversion server-side if `filter_args` is provided. Fortunately `_rangeindex_to_multiindex()` is already written to detect if it has already been run. That was implemented to make sure that it still runs when using updated clients with outdated servers, but avoid running it twice when updated clients request from updated servers. One small difference though would be that `df.sort_index()` won't be run before `_extract_n_sequences_from_df()` if  `_rangeindex_to_multiindex()` isn't called server-side. That would affect which rows are returned if they are added to lyse out of order, e.g. if a user adds files to it manually using the "+" button. I actually just moved `df.sort_index()` from `data()` to `_rangeindex_to_multiindex()` in 4c944dd for that very reason.

zakv: Actually,  I took a look at the source code and it seems that `set_index(inplace=False)` [calls `self.copy()`](https://github.com/pandas-dev/pandas/blob/db08276bc116c438d3fdee492026f8223584c477/pandas/core/frame.py#L4553-L4556) which [makes a deep copy](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.copy.html) by default.

There's still a chance that it's a copy-on-write but from what I've read I get them impression that copy-on-write is on some pandas todo/wishlists but isn't implemented. So I'm becoming more confident that it always does do a deep copy immediately.

zakv: I'm wondering how resource-heavy making the copy actually is though. I have an instance of lyse running at the moment with ~3,000 shots in it. Requesting the dataframe takes ~0.5 seconds, but running `df = lyse.data(); df.memory_usage(deep=True).sum()` shows that it's still only 15 MB. Also, running `df.copy(deep=True)` only takes about 7 ms. That's not bad, and it still wouldn't be terrible with ~10x more shots assuming things scale linearly.

lyse/__init__.py
zakv: Added in 8d2a69e.

zakv:
Looks like some PRs got merged into master. Should I do a rebase / force push for this branch? Or is the plan to just do a merge?

philipstarkey:
> Looks like some PRs got merged into master. Should I do a rebase / force push for this branch? Or is the plan to just do a merge?

I don't think you need to rebase unless you want to test the docs build. The rebase will probably also stuff up the git commit references in your comments, so I'd probably lean towards just merging and then patching any issues with the docs separately.

lyse/__main__.py
philipstarkey: Ok, so I wrote something (below the horizontal rule) about more tests (left for posterity) but I don't know whether they are important to do.

If copying is always 100 times faster than transmitting, then there is very little need to optimise for the case where no filtering/slicing is done (which is I guess what my suggestion boils down to). Could you do a brief test to see how `df = lyse.data();` vs `df.copy(deep=True)` scales with more shots and more columns? There will be some overhead for small numbers of shots (meaning `lyse.data` will stay ~constant while `copy` grows), but at some point I would expect them to grow with some constant ratio. If we confirm that ratio is big (like ~100 as you saw for 3000 shots), then we should always do the copy on the server side, and only have that run in the main loop (probably explicitly do `copy(deep=True)` in mainthread and then do an inplace index modification on the copy in the zmqserver thread). The question is whether 3000 shots is "big" or not!

-----
Hmm, interesting.

Could you do a couple of other checks on the performance of `copy(deep=True)`?

1. Can you confirm it does scale approximately linearly with the number of shots?
2. Could you report the copy performance with a df with those 3000 shots in and then with an additional 3000 shots that have had a different analysis script run on them (this makes the dataframe expand columns and fill in NaNs for the new columns, old shots intersection)?

Also, couple of questions about the shots:

1. How many globals are in the shots?
2. How many analysis result columns are in the dataframe?

zakv: Good questions. I'm still working on answering some of them, but I'll respond with what I have now before I lose track of it.

> 1. How many globals are in the shots?

About 250.

> 2. How many analysis result columns are in the dataframe?

About 25.

> Can you confirm it (`copy(deep=True)`) does scale approximately linearly with the number of shots?

Running the following code:

```python
from time import perf_counter
import matplotlib.pyplot as plt
import numpy as np
import lyse

df = lyse.data()  # Need at least 3000 runs in dataframe for this test.

times = []
sizes = np.linspace(1, 3001, 31)

for size in sizes:
    size = int(size)
    df_subset = df.iloc[:size]
    start = perf_counter()
    df_subset.copy(deep=True)
    end = perf_counter()
    times.append(end-start)

fig = plt.figure()
axes = fig.add_subplot(111)
axes.plot(
    sizes,
    times,
    marker = 'o',
)
axes.set_title('Dataframe Deep Copy Times')
axes.set_xlabel('Number of Rows')
axes.set_ylabel('Duration (s)')
```

Gave the following plot:

image

Which looks pretty linear.

zakv: > 2 Could you report the copy performance with a df with those 3000 shots in and then with an additional 3000 shots that have had a different analysis script run on them (this makes the dataframe expand columns and fill in NaNs for the new columns, old shots intersection)?

I figured it would take a while to set up that exact test, so I did something similar but slightly different. I used the same dataframe with ~3000 shots, but used the "Mark as not done" feature to run a singleshot lyse script on only ~5 of those runs. That produced 10 results which weren't calculated for the other runs so the dataframe had a lot of empty entries. I ran the singleshot routine on some rows around index 10 in the dataframe so that they would be included in most of the subsets. Re-running that code from the last comment gave the plot below:

image

So the results are pretty similar with several extra columns that have a lot of empty entries.

I'll set up a script to time `df = lyse.data()` and `df.copy(deep=True)` as the size of the dataframe changes while I run an M-LOOP optimization. That'll take a while to complete but hopefully I'll have something by the end of the day.

zakv: > If copying is always 100 times faster than transmitting, then there is very little need to optimise for the case where no filtering/slicing is done (which is I guess what my suggestion boils down to). Could you do a brief test to see how df = lyse.data(); vs df.copy(deep=True) scales with more shots and more columns? 

This only goes up to ~1700 shots but the scaling is pretty clear. It seems that the time for `data()` is always roughly equal to 100 times the time required for `copy()`.

image

I'm not sure why the yellow curve jumps at ~500 entries. My best guess is that it's some internal pandas thing about how small vs big dataframes are copied. At any rate, after that point the two curves scale very similarly.

I should also mention that all of these timing tests were done with a copy of lyse that doesn't include the changes in this PR.

lyse/__init__.py
```diff
@@ -76,28 +76,133 @@ class _RoutineStorage(object):
...
+
+    Raises:
+        ValueError: If `n_sequences` isn't `None` or a nonnegative integer, then
+            a `ValueError` is raised. Note that no `ValueError` is raised if
+            `n_sequences` is greater than the number of sequences available. In
+            that case as all available sequences are returned, i.e. the entire
+            lyse dataframe is returned.
+
+    Returns:
+        (pandas series or dataframe): If `filepath` is provided, then a pandas
```
philipstarkey, line 139: In trying this in my own docs, I was unable to see the return type if it was surrounded in brackets.

Updating it to ``:obj:`pandas:pandas.DataFrame` or :obj:`pandas:pandas.Series`: If `filepath` is provided, then a pandas`` has the intended effect

zakv: Thanks for figuring that out, updated in bff7d19

philipstarkey: According to the style guide, you only need to indent from the `Returns` statement. Multiline return text can all be at the same level, even when specifying a return type (you've indented subsequent lines after the return type as well). I have no idea if this matters or not!

lyse/__main__.py
philipstarkey: Ok. Think I'm happy with this. We can revisit if someone complains about performance.

lyse/__init__.py
zakv: Good catch, corrected in f80f182.

Issue #37, opened 2018-03-01
Specify formatting/existence of print statement in save_results


philipstarkey:
**Original report ([archived issue](https://labscript-suite.github.io/labscript-suite-bitbucket-archive/#!/labscript_suite/lyse/issues/37)) by Russell Anderson (Bitbucket: [rpanderson](https://bitbucket.org/%7B6c12ac6e-1980-466d-86fe-6c163bd116c2%7D/), GitHub: [rpanderson](https://github.com/rpanderson)).**

----------------------------------------
When saving analysis results to the shot file, [`save_results`](https://github.com/labscript-suite/lyse/blob/4755cce3e4a0189da298899fffb13c8067258e7c/__init__.py#L233) prints the name and value without formatting, e.g.
~~~~
saving MOT_load_rate = 509343215.257
saving MOT_number = 2574843222.31
saving MOT_beam_power = 118.185460376
saving u_MOT_beam_power = 0.799789946506
~~~~
This enhancement would permit the value to be formatted using a format-string (perhaps honouring uncertainties), or permit the print statement to be omitted altogether. For example:
~~~~
saving MOT_load_rate = 5.09e8
saving MOT_number = 2.575e9
saving MOT_beam_power = 118.2(8)
~~~~

philipstarkey:
Closing in favour of #84

PR #80, opened 2020-11-02
Fix 78

### LINKED ISSUE #78: Run.set_group behaviour with no_write flag
There is potential unexpected behaviour with a call to `Run.set_group` if `no_write` was explicitly set to `True` when instantiating the `Run` object. This is consistent with the behaviour prior to #71, but is perhaps not consistent with the intention.

#72 fixed a bug with the implementation of #71, but using the `no_write=True` option at instantiation might mean that the code still raises an exception since the results group may not exist for the subgroup to be created in. #72 also modifies the file regardless of the `no_write` state.

#71 attempted to back up/restore the `no_write` flag state, but I'm not sure if this even made sense. Should `set_group()` mean that you intent to write to the file? Whatever behaviour we settle on, this method needs documenting!

zakv:
This PR attempts to resolve #78 

Changes:
* Prevent `Run.set_group()` from creating the group in the hdf5 file if `self.no_write` is `True`.
* Make `Run._create_group_if_not_exists()` raise an error if the group doesn't already exist but `self.no_write` is `True`.
* `Run._create_group_if_not_exists()` now catches the `ValueError` raised if something else creates the group between when  `Run._create_group_if_not_exists()` sees that the group doesn't exist and when it actually tries to create the group.
* Ditch the usage of `self._no_group` to backup/restore `self.no_write`.
  * Instead set `self.group` to `None` by default if there's no lyse script running to name it after, then set the save methods to raise an exception if they need `self.group` but it is `None`.
* Removed the `print()` statement from `Run.save_results()`.
* Make `Run`'s `h5_path`, `no_write`, and `group` attributes private properties.
* Make error messages in `Run.save_result()` and `Run.save_result_array()` more accurate.
* Resolve the bug where `Sequence.__init__()` would fail if the hdf5 file didn't already exist, first brought up in PR #73 .
* Refactor `Sequence.__init__()` to use `Run.__init__()`.
* Add some docstrings in `__init__.py`.
* Fix a few typos in other parts of the code.

### DIFF
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -116,9 +116,9 @@ class Run(object):
             to the file. Defaults to `False`.
     """
     def __init__(self,h5_path,no_write=False):
-        self.__h5_path = h5_path
-        self.__no_write = no_write
-        self.__group = None
+        self._h5_path = h5_path
+        self._no_write = no_write
+        self._group = None
         if not self.no_write:
             self._create_group_if_not_exists(h5_path, '/', 'results')
                      
@@ -142,12 +142,12 @@ class Run(object):
     @property
     def h5_path(self):
         """str: The value provided for `h5_path` during instantiation."""
-        return self.__h5_path
+        return self._h5_path
 
     @property
     def no_write(self):
         """bool: The value provided for `no_write` during instantiation."""
-        return self.__no_write
+        return self._no_write
 
     @property
     def group(self):
@@ -164,7 +164,7 @@ class Run(object):
             Attempting to directly set `self.group`'s value will automatically
             call `self.set_group()`.
         """
-        return self.__group
+        return self._group
 
     @group.setter
     def group(self, value):
@@ -206,7 +206,7 @@ class Run(object):
                 `'/results'` group of the hdf5 file.
         """
         self._create_group_if_not_exists(self.h5_path, '/results', groupname)
-        self.__group = groupname
+        self._group = groupname
 
     def trace_names(self):
         with h5py.File(self.h5_path, 'r') as h5_file:
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -238,10 +238,43 @@ class Run(object):
         return results        
             
     def save_result(self, name, value, group=None, overwrite=True):
-        """Save a result to h5 file. Defaults are to save to the active group 
-        in the 'results' group and overwrite an existing result.
-        Note that the result is saved as an attribute of 'results/group' and
-        overwriting attributes causes h5 file size bloat."""
+        """Save a result to the hdf5 file.
+
+        With the default argument values this method saves to `self.group` in
+        the `'/results'` group and overwrites any existing value. Note that the
+        result is saved as an attribute and overwriting attributes causes hdf5
+        file size bloat.
+
+        Args:
+            name (str): The name of the result. This will be the name of the
+                attribute added to the hdf5 file's group.
+            value (any): The value of the result, which will be saved as the
+                value of the hdf5 group's attribute set by `name`. However note
+                that when saving large arrays, it is better to use the
+                `self.save_result_array()` method which will store the results
+                as a dataset in the hdf5 file.
+            group (str, optional): The group in the hdf5 file to which the
+                result will be saved as an attribute. If set to `None`, then the
+                result will be saved to `self.group` in `'/results'`. Note that
+                if a value is passed for `group` here then it will NOT have
+                `'/result'` prepended to it. This is in contrast to using the
+                default group set with `self.set_group()`; when the default
+                group is set with that method it WILL have `'/results'`
+                prepended to it when before saving results. Defaults to `None`.
+            overwrite (bool, optional): Sets whether or not to overwrite the
+                previous value if the attribute already exists. If set to
+                `False` and the attribute already exists, an `Exception` is
+                raised. Defaults to `True`.
+
+        Raises:
+            Exception: An `Exception` is raised if `self.no_write` is `True`
+                because saving the result would edit the file.
+            ValueError: A `ValueError` is raised if `self.group` is `None` and
+                no value is provided for `group` because the method then doesn't
+                know where to save the result.
+            Exception: An `Exception` is raised if an attribute with name `name`
+                already exists but `overwrite` is set to `False`.
+        """
         if self.no_write:
             raise Exception('This run is read-only. '
                             'You can\'t save results to runs through a '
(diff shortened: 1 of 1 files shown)


zakv:
A few other points that I was hoping for feedback on before making any changes:

Is the print statement here:
https://github.com/labscript-suite/lyse/blob/091044719c801972f62a78fc43e9ae22bd5d0461/lyse/__init__.py#L277-L286
intended to be there because it's somewhat easy to accidentally call the method incorrectly? Or was that a debugging print statement that got left in accidentally?

Also, as currently written, the `group` argument for the `save...()` methods doesn't get `'/results/'` prepended to it automatically. That means that to save to e.g. `'/results/results_group`' users either have to call `run.set_group('results_group')` or `run.save_result('..., group='/results/results_group')`. It would be easy to forget to manually prepend `'/results/'` in the latter. Should I adjust the saving methods to automatically prepend `'/results/'` to the `group` argument? I suppose that would technically break backwards compatibility, though I'm not sure if many people use that argument.

philipstarkey:
So the print statement has been there for over 9 years (since before it was even named lyse - which is as far back as I can check). I would nuke it provided you add a docstring to that method (so that users have a way to not be confused!)

I think the lack or `'/results/'` prepend is so that you can save results anywhere in the file if you want. This was probably introduced for Monash labs who wanted to overwrite the images saved by BIAS? Either way I think it has to stay as is. You should be calling `set_group()` if you are running from a terminal (running from a file negates the need for an explicit call to `set_group()`). The `group` argument of the `save*` methods should only be used if you want to save to an arbitrary location (and future docstrings should be written to reflect that).

philipstarkey:
Couple of other suggested changes:

1. This one is technically backwards incompatible, but I think we should just do it anyway. I would rename `self.no_write` to `self.__no_write` so that it can't be changed after instantiation. This variable should never have been public, so I would treat the rename as a bug fix so we don't have to bump the major version. To be honest..., h5_path and group should be private too. Gah, this is a bit of scope creep, but do you want to rename those variables to `self.__...` and replace all 3 of them with read only Python properties with the original names so that backwards compatibility is partially maintained? Reading the values is fine, being able to change them is a bug.

2. There are some exceptions raised in `save_result_array` and `save_result` which sort of imply that `no_write` will only be set when using a `Sequence` object. This isn't really true - `Run` objects could be deliberately created in read only mode by the user.

3. The first line of the `Run` docstring uses the word "run" when you mean "shot". This is of course because the class name is wrong (historical holdover from before we had settled on shot). At the very least we should make the docstring use "shot" instead of "run". That would also necessitate updating a few of the error messages to use the word shot instead of run. I'm going to log a bug for renaming and deprecating `Run` while I'm at it.

4. Does this interact with PR #73? It was a minimal change proposed there...do you want to integrate it into yours (see my comments on that PR for changes) and update to be consistent with the current `no_write` changes?

Think that's everything from me! Sorry it's a bit of scope creep.

zakv:
Ok! I believe that last batch of commits resolves all of the issues you've mentioned, including the bug that PR #73 aims to fix.

> I think the lack or '/results/' prepend is so that you can save results anywhere in the file if you want.

Ah, I had assumed that using `'..'` would let you easily access the root directory from `/results` but seems I was wrong. It definitely makes sense to leave it as is.

lyse/__init__.py
```diff
@@ -102,35 +102,74 @@ def globals_diff(run1, run2, group=None):
...
+            file for a shot.
+        no_write (bool, optional): Set to `True` to prevent editing the shot's
+            hdf5 file. Note that doing so prohibits the ability to save results
+            to the file. Defaults to `False`.
+    """
     def __init__(self,h5_path,no_write=False):
-        self.no_write = no_write
-        self._no_group = None
-        self.h5_path = h5_path
+        self._h5_path = h5_path
```
philipstarkey, line 119: These should have double underscores to make them private (but see note on `Sequence` class before changing

lyse/__init__.py
```diff
@@ -418,32 +578,42 @@ def globals_diff(self, other_run, group=None):
         
 class Sequence(Run):
     def __init__(self, h5_path, run_paths, no_write=False):
```
philipstarkey, line 580: A significant amount of this `__init__` method appears to overlap with the base class. There are minor differences (like `f_locals` vs `f_globals` but I'm not sure there are material differences. I also suspect that making the instance attributes private (see other comment in this review) may prevent the properties of the base class from working in the child class (could be wrong about this). However, calling `super().__init__` in the child class and having the parent set those values would negate the issue (should it exist).

Can you see any reason why `Sequence.__init__` isn't just an extension of `Run.__init__` provided you attempt to create the file if it doesn't exist before calling the parent method?

philipstarkey (commented):
Looking good! Couple more review comments for your latest changes.

lyse/__init__.py
zakv: Yeah, I agree that refactoring the `__init__()` methods would have been nice. It's possible but I couldn't get it to be pretty so I abandoned the idea and went with single underscore private attributes. The issue has to do with getting the name of the lyse script.

When calling `Run.__init__()` directly from a lyse script, both `f_globals['__file__']` and `f_locals['__file__']` are the path to the script. However things go wrong when the lyse script calls `Sequence.__init__()` then that calls `Run.__init__()` which then tries to determine its calling script's name. In that case `f_globals['__file__']` gets set to the path to lyse's `__init__.py` and `f_locals` doesn't have a `__file__` key. That meant either `self.group` would be set to `__init__` if using `f_globals` or using `f_locals` would raise a `KeyError` which would then be suppressed.

As a workaround for that I added an optional `group` argument to `Run.__init__()` which defaulted to `None` and would be automatically determined from `f_globals` if it was `None`, so directly calling `Run.__init__()` worked as before. I then had `Sequence.__init__()`  evaluate `f_globals['__file__']` itself, which gave the name of the lyse script, and pass that value to `Run.__init__()`.

That mostly worked, but creating a `Sequence` from outside of lyse had issues. In that case `Sequence.__init__()` would correctly fail to get a name of a lyse script and then would pass `group=None` to `Run.__init__()`. However that would make `Run.__init__()` try to automatically determine the calling script, which it would think was lyse's `__init__.py`, so it would set `self.group` to `__init__` instead of `None`.

I can think of two ways to work around that last issue and make instantiating a `Sequence` from an interpreter work. One option is to have `Sequence.__init__()` pass some other special non-default value for `group` to `Run.__init__()` which would tell it not to try to automatically determine `self.group` and instead just set it to `None`. Another option is to have `Run.__init__()` use `f_locals` instead of `f_globals` so that it fails to automatically get a script name when `Sequence.__init__()` calls it with `group=None`. I can't tell if that's hacky or clever. It definitely makes the code hard to understand though since you have know about some intricacies of how the values in `f_locals` depend on the caller.

Given that the first workaround didn't seem appealing I ended up sticking with the lyse's previous approach with duplicated code and used single underscore attributes. I didn't think of the second workaround until writing this, but I'm not sold on it either to be honest.  Maybe there's a better option?

philipstarkey: Ah I see. I'll see if I can determine a better workaround!

philipstarkey: Ok, so I think that the solution might be to pass in `frame=inspect.currentframe().f_back` as an optional argument to `Run`. I suspect that the issue is this (inside `Run.__init__` points to the `Sequence.__init__` method and you would need an additional `f_back` chained on to get out of `Sequence` and back into the lyse script). So obtaining that value from within `Sequence` and passing it to `Run` should resolve the issue I think.

I might be wrong though, because that would imply using a `Run` object from within an imported function would result in the wrong group name being set. But maybe no one has ever tried instantiating a `Run` from within an imported module? If that does result in a bug, then we actually need to inspect the whole stack until we get up to `os.path.join(lyse.LYSE_DIR,'analysis_subprocess.py')` and then go down one level to get the lyse script name (I think?). This solution (if it works) would also solve the problem and would not need the additional kwarg (since it would walk to stack until it finds the right value).

zakv: Nice! That second idea worked great and is implemented in 3d38b3b.

lyse/__init__.py
zakv: Updated in 7ba7062 after making the changes suggested below.

lyse/__init__.py
philipstarkey: Great! Last nitpick: does your current implementation fail if the analysis script itself is also called analysis_subprocess.py?

zakv: Yeah I believe it would. I'll adjust it to use your `os.path.join....` suggestion to be safe.

zakv: Updated in 1dac68f.

zakv:
Also, I just noticed issue #37. This PR sort of negates that issue since it removes that print statement.

philipstarkey:
Cool! I will merge this.

I've also logged #83 which is sort of related, but too hard to fix here since it is historic behaviour. I'll probably close #37 in favour of #84 as well.

Issue #81, opened 2020-11-03
Rename Run to Shot


philipstarkey:
The `Run` class is poorly named. It should really be `Shot`. I propose we rename, and add the following class for backwards compatibility until v4.0

```
import warnings
class Run(Shot):
    def __init__(self, *args, **kwargs):
        warnings.warn("The 'Run' class has been renamed to 'Shot' (but is otherwise compatible). 'Run' will be removed in lyse v4.0. Please update your analysis code to use the 'Shot` class", DeprecationWarning)
        super().__init__(*args, **kwargs)
```

chrisjbillington:
Very yes.

Runmanager -> shotmanager? More entrenched, maybe :p

philipstarkey:
> Runmanager -> shotmanager? More entrenched, maybe :p
 
Paramanager? Paramager?

(no I'm not being serious!)

PR #85, opened 2020-11-05
Renaming Run to Shot

philipstarkey:
Resolves Rename Run to Shot #81

Renamed classes/arguments/docs from run to shot and added dprecation notices where appropriate.

Arguments (distinct from keyword arguments) were renamed without worrying about backwards compatibility under the assumption that people are not using them as keyword arguments (which is technically possible).

### DIFF
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -13,12 +13,14 @@
 
 from lyse.dataframe_utilities import get_series_from_shot as _get_singleshot
 from labscript_utils.dict_diff import dict_diff
+import copy
 import os
 import socket
 import pickle as pickle
 import inspect
 import sys
 import threading
+import warnings
 
 import labscript_utils.h5_lock, h5py
 from labscript_utils.labconfig import LabConfig
@@ -79,7 +81,7 @@ routine_storage = _RoutineStorage()
 def data(filepath=None, host='localhost', port=_lyse_port, timeout=5, n_sequences=None, filter_kwargs=None):
     """Get data from the lyse dataframe or a file.
     
-    This function allows for either extracting information from a run's hdf5
+    This function allows for either extracting information from a shots's hdf5
     file, or retrieving data from lyse's dataframe. If `filepath` is provided
     then data will be read from that file and returned as a pandas series. If
     `filepath` is not provided then the dataframe in lyse, or a portion of it,
@@ -97,11 +99,11 @@ def data(filepath=None, host='localhost', port=_lyse_port, timeout=5, n_sequence
     that method.
 
     Args:
-        filepath (str, optional): The path to a run's hdf5 file. If a value
+        filepath (str, optional): The path to a shot's hdf5 file. If a value
             other than `None` is provided, then this function will return a
-            pandas series containing data associated with the run. In particular
+            pandas series containing data associated with the shot. In particular
             it will contain the globals, singleshot results, multishot results,
-            etc. that would appear in the run's row in the Lyse dataframe, but
+            etc. that would appear in the shot's row in the Lyse dataframe, but
             the values will be read from the file rather than extracted from the
             lyse dataframe. If `filepath` is `None, then this function will
             instead return a section of the lyse dataframe. Note that if
@@ -203,10 +205,10 @@ def _rangeindex_to_multiindex(df, inplace):
     df.sort_index(inplace=True)
     return df
 
-def globals_diff(run1, run2, group=None):
-    return dict_diff(run1.get_globals(group), run2.get_globals(group))
+def globals_diff(shot1, shot2, group=None):
+    return dict_diff(shot1.get_globals(group), shot2.get_globals(group))
  
-class Run(object):
+class Shot(object):
     """A class for saving/retrieving data to/from a shot's hdf5 file.
 
     This class implements methods that allow the user to retrieve data from a
@@ -229,9 +231,9 @@ class Run(object):
                      
         try:
             if not self.no_write:
-                # The group where this run's results will be stored in the h5
+                # The group where this shot's results will be stored in the h5
                 # file will be the name of the python script which is
-                # instantiating this Run object. Iterate from innermost caller
+                # instantiating this Shot object. Iterate from innermost caller
                 # to outermost. The name of the script will be one frame in
                 # from analysis_subprocess.py.
                 analysis_subprocess_path = os.path.join(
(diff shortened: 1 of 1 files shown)


philipstarkey:
@zakv Would also appreciate your eyes on this (don't seem to be able to explicitly request a review from you using the GitHub interface since you are not a project member. Might have to fix that)

lyse/__init__.py
```diff
@@ -724,15 +735,23 @@ def __init__(self, h5_path, run_paths, no_write=False):
...
+            shot_paths = shot_paths['filepath']      
+        self.__shots = {path: Shot(path,no_write=True) for path in shot_paths}
+        
+    @property
+    def runs(self):
+        warnings.warn("The 'runs' attribute has been renamed to 'shots'. 'runs' will be removed in lyse v4.0. Please update your analysis code to use the 'shots` attribute", DeprecationWarning)
+        return copy.copy(self.__shots)
+
+    def shots(self):
+        return copy.copy(self.__shots)
```
zakv, line 748: I believe this needs a property decorator

```suggestion
    @property
    def shots(self):
        return copy.copy(self.__shots)
```

lyse/__init__.py
```diff
@@ -703,12 +705,21 @@ def globals_groups(self):
...
-class Sequence(Run):
-    def __init__(self, h5_path, run_paths, no_write=False):
+
+class Run(Shot):
+    def __init__(self, *args, **kwargs):
+        warnings.warn("The 'Run' class has been renamed to 'Shot' (but is otherwise compatible). 'Run' will be removed in lyse v4.0. Please update your analysis code to use the 'Shot` class", DeprecationWarning)
+        super().__init__(*args, **kwargs)
+
+    def globals_diff(self, other_run, group=None):
+        return globals_diff(self, other_run, group)
```
zakv, line 718: I think the inherited version of this method would work fine right? Is this included just in case users had explicitly called this method with `other_run=...`, using it as a named argument rather than a positional one?

lyse/__init__.py
```diff
@@ -703,12 +705,21 @@ def globals_groups(self):
...
+    def globals_diff(self, other_shot, group=None):
+        return globals_diff(self, other_shot, group)            
     
-        
-class Sequence(Run):
-    def __init__(self, h5_path, run_paths, no_write=False):
+
+class Run(Shot):
+    def __init__(self, *args, **kwargs):
+        warnings.warn("The 'Run' class has been renamed to 'Shot' (but is otherwise compatible). 'Run' will be removed in lyse v4.0. Please update your analysis code to use the 'Shot` class", DeprecationWarning)
```
zakv, line 714: This line is pretty long; maybe use `dedent()`. Also it looks like the last occurrence of the word 'Shot' has one single quote and one backtick instead of two single quotes, and the second sentence is missing a period. Hope that's not too obnoxious and nitpicky.

zakv (commented):
Overall looks good! I made a few comments, and have a couple of overall questions.

1. Is there any particular reason for using double underscores rather than single? It's not clear to me how the name mangling helps.
1. Why is "Shot" better than "Run"? I agree that normally I'd use the word "shot" (in fact our lab uses a custom `Shot` class than inherits from `lyse.Run`) but I would think that "run" would mean the same thing. Is the word "run" misleading somehow? I agree that if `lyse` were being written from scratch I'd rather name this class `Shot` than `Run` but I'm not 100% convinced it's confusing enough to be worth making users update all of their analysis scripts and other code.

zakv:
Also, `Examples.rst` needs to be updated to use `Shot` instead of `Run`, and `index.rst` needs one line updated ("the entire dataset from all the runs that are currently loaded").

lyse/__init__.py
philipstarkey: Yeah, that was the reasoning. Unlikely (and I argued above unnecessary) but the opportunity to do it was available and it's more likely that someone is using this method than the underlying `globals_diff` function.

philipstarkey:
> Overall looks good! I made a few comments, and have a couple of overall questions.
> 
> 1. Is there any particular reason for using double underscores rather than single? It's not clear to me how the name mangling helps.

In this instance I did it to made the variable read only. Not strictly necessary, but there it's really an internal variable and it's better practice to keep it from being modified outside of the class. It's basically just defensive programming, and makes it easier for internal changes to be made down the line without worrying about whether we're breaking the API.

> 2. Why is "Shot" better than "Run"? I agree that normally I'd use the word "shot" (in fact our lab uses a custom `Shot` class than inherits from `lyse.Run`) but I would think that "run" would mean the same thing. Is the word "run" misleading somehow? I agree that if `lyse` were being written from scratch I'd rather name this class `Shot` than `Run` but I'm not 100% convinced it's confusing enough to be worth making users update all of their analysis scripts and other code.

Run could be interpreted as a set of shots (a "run" of shots"). Run was very early nomenclature we adopted, but it rapidly become confusing when used in every day speech. We thus settled towards the more obvious "Shot" and "Sequence" which cannot be confused. We were hesitant to change the API at the time, but I think it's probably time to bite the bullet and do it. I agree it's a bit of a pain to change code, but `lyse.Run` is not going anywhere for a long time (can't imagine lyse 4.0 is close given 3.0 was only just released) and it's a 5 second fix per script if you want the deprecation warning to go away. I could maybe make it so the deprecation warning only showed up once on the first shot and not subsequent shots? But that risks being missed if users don't restart lyse often.

lyse/__init__.py
philipstarkey: Thanks. This is perfect (and now fixed in 016768d)

lyse/__init__.py
philipstarkey: Fixed in 016768d

zakv:
> It's basically just defensive programming, and makes it easier for internal changes to be made down the line without worrying about whether we're breaking the API.

Good to know, thanks! I wasn't sure if there was something that the double underscore name mangling signaled to users that a single underscore didn't.

> We thus settled towards the more obvious "Shot" and "Sequence" which cannot be confused

To be honest our lab used to use "shot" and "sequence" interchangeably, haha. That was back when we used a labview-based control system where you designed shots by listing a sequence of output times/voltages in a table. I suspect though that the large majority of people would interpret "shot" and "sequence" in the way you mentioned.

All your changes look good from my perspective. The PR looks ready to merge as far as I can tell.

PR #87, opened 2020-11-07
Fix 86

### LINKED ISSUE #86: Run.get_units() doesn't use its group argument
Just noticed this while looking at PR #85. Should be an easy fix.

zakv:
This PR resolves #86 by making `Run.get_units()` only return units for the group specified by its optional `group` argument. If it is left at its default value of `None`, then the units for all globals in all groups are returned as before. Also, a docstring was added for that method.

### DIFF
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -692,8 +692,15 @@ class Run(object):
                 temp_dict = dict(obj.attrs)
                 for key, val in temp_dict.items():
                     units_dict[key] = val
-        with h5py.File(self.h5_path, 'r') as h5_file:
-            h5_file['globals'].visititems(append_units)
+        if group:
+            try:
+                with h5py.File(self.h5_path, 'r') as h5_file:
+                    h5_file['globals'][group].visititems(append_units)
+            except KeyError:
+                pass
+        else:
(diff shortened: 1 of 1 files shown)


philipstarkey:
Wondering if the following is a cleaner implementation (the `append_units` function seems overly complex for what it does).

```
    def get_units(self, group=None):
        units = {}
        path = 'globals'
        if group is not None:
            path += f'/{group}'
        with h5py.File(self.h5_path, 'r') as h5_file:
            h5_file[path].visititems(
                lambda n, o: units.update(dict(o.attrs) if 'units' in n else {})
            )
        return units
```
(This has not been tested. I'm not even sure if that `dict` above is necessary either)

There are two main changes in this:
1. Dynamically constructing path rather than using if/else
2. Ditching the function for a lambda.

The first change is probably cleaner, but second might not be worth it. Could instead use:
```
        def append_units(name, obj):
            if 'units' in name:
                units.update(dict(obj.attrs))
```
(I think)

zakv:
Yeah you're right that method could be simplified a bit. I still went with `append_units()` over the lambda function just because I thought it made the code a bit easier to read/understand, though maybe a bit less pythonic.

PR #88, opened 2020-11-19
Move infer_objects() out of main thread

zakv:
Previously when `lyse.data()` was called, the server would call `infer_objects()` and `copy()` in the main thread on the lyse dataframe. It turns out that `infer_objects()` actually takes a few times longer than `copy()` and it doesn't need to happen in the main thread. This PR moves the call to `infer_objects()` to outside of the main thread to reduce the main thread's workload.

This change does mean that `infer_objects()` is never called on the original lyse dataframe; it's only called on deep copies of it. As far as I can tell this doesn't cause any issues though.

### DIFF
--- a/lyse/__main__.py
+++ b/lyse/__main__.py
@@ -188,14 +188,18 @@ class WebServer(ZMQServer):
                 "'get dataframe'\n 'hello'\n {'filepath': <some_h5_filepath>}")
 
     @inmain_decorator(wait_for_return=True)
+    def _copy_dataframe(self):
+        df = app.filebox.shots_model.dataframe.copy(deep=True)
+        return df
+
     def _retrieve_dataframe(self):
         # infer_objects() picks fixed datatypes for columns that are compatible with
         # fixed datatypes, dramatically speeding up pickling. It is called here
         # rather than when updating the dataframe as calling it during updating may
         # call it needlessly often, whereas it only needs to be called prior to
         # sending the dataframe to a client requesting it, as we're doing now.
-        app.filebox.shots_model.infer_objects()
-        df = app.filebox.shots_model.dataframe.copy(deep=True)
+        df = self._copy_dataframe()
+        df.infer_objects()
         return df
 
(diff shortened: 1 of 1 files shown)


zakv:
Also, I was worried that only ever calling `infer_objects()` on copies of the dataframe would be slow because it would have to be done from scratch each time. This doesn't seem to be the case though.

I logged the timings of various steps associated with `lyse.data()` and the execution time of `infer_objects()` was similar whether it was always run in the main thread on the same original dataframe or whether it was only ever run outside the main thread on copies of the dataframe. With 10,000 shots in the dataframe, `infer_objects()` took about 105 ms when run only in the main thread and it took about 70 ms when only run outside the main thread. For comparison, the call to `copy()` took about 25 ms in either case.

dihm:
Hi Zak, I'm looking to clear out this older PR. It seems innocent enough, but I'm curious what the underlying motivation is for it. Were you finding your main thread to be overloaded somehow?

zakv:
I'm having a hard time remembering since this was a fair while ago. I believe this was inspired by the discussion [here](https://github.com/labscript-suite/lyse/pull/76#discussion_r514875633). Before that PR, the lyse GUI would freeze up for a noticeable amount of time (~seconds) when requesting the dataframe if there were ~1000 shots in the dataframe. To avoid that, much of the heavy lifting was moved out of the main lyse thread. Instead the main thread just makes a deep copy of the dataframe, then a separate thread handles the rest of the work of shipping the dataframe out over ZMQ. It turns out that copying the dataframe actually happens pretty quickly, so with this setup the GUI remains much more responsive.

This PR is a further effort along that same line, trying to move as much work out of the main thread as possible to minimize how long the GUI becomes unresponsive. I couldn't tell you off the top of my head how much this PR helped in an absolute sense (i.e. made the GUI become responsive again x seconds faster when the dataframe has y shots), but from the first comment in this PR it seems like it did speed things up.

Issue #82, opened 2020-11-05
df.to_msgpack and pandas.read_msgpack removed from pandas


zakv:
Pandas [dropped support for msgpack](https://github.com/pandas-dev/pandas/issues/27722), I believe in the 0.25 to 1.0 transition judging by comparing different tagged versions [here](https://github.com/pandas-dev/pandas/blob/1.0.x/pandas/core/generic.py. ). Unfortunately this breaks lyse's "save dataframe" and "load dataframe" features.

Maybe it's best to switch to `df.to_pickle()` and `pandas.read_pickle()`? I don't see any good ways to support loading dataframes from old `msgpack` saves though, except maybe through some third party library.

lakodarmstadt:
I think using pandas.to_parquet is an option. I tried pandas.to_feather first but this has difficulties with the multiindex structure of our dataframes. For pandas >=1.2 this is solved for pandas.to_parquet.
How can I set the requirement for pandas>=1.2? I can use the requirement textfile but this is only used during installation if I understand correctly.

dihm:
Sorry for being very late to the party here, but I'm finally getting to grips with the issue here. I'm a little concerned about using parquet. I just tried to test #91. Not only does it require an optional dependency with lots of its own dependenices (pyarrow or fastparquet+numba), but the current pyarrow conda package is bugged and doesn't even work (you have to install via pip, see AnacondaRecipes/pyarrow-feedstock#2).

For now, I'm more inclined to use the pickle interface (with or without compression) as a more portable, working out of the box solution. Based on [this random blog post from years ago](http://matthewrocklin.com/blog/work/2015/03/16/Fast-Serialization), it seems pickle isn't that off in terms of performance either.

As for backwards compatibility, I'm not really sure what to do. There is a [third party package](https://github.com/TyberiusPrime/pandas-msgpack), but support seems extremely limited. Perhaps the best solution is to just not support loading old msgpacked files and telling the user to either manually convert to pickle (using 3rd party resources or an older version of pandas) or reload the source files and resave the dataframes as pickle files.

Does anyone have any thoughts?

zakv:
Yeah I don't think there's a good way to handle this. Either we add a bunch of dependencies or we lose the ability to load old dataframes. My vote is to lose the ability to load old dataframes.

[...]

PR #91, opened 2021-03-16
Use to_parquet instead of depricated msgpack

lakodarmstadt:
This solves issue 82 (https://github.com/labscript-suite/lyse/issues/82) by offering a new possibility to save Dataframes. I am still unsure how to ensure pandas<1.2.
Furthermore I solve two bugs that came up:
1. The first shot in the saved dataframe was always reloaded, even if it had not changed.
2. The progress bar was stuck after loading a dataframe where no files had changed. Maybe the bugfix I used is not the best ... I am open for better ideas.

### DIFF
--- a/lyse/__main__.py
+++ b/lyse/__main__.py
@@ -2243,23 +2243,23 @@ class Lyse(object):
             for sequence in sequences:
                 sequence_df = pandas.DataFrame(df[df['sequence'] == sequence], columns=df.columns).dropna(axis=1, how='all')
                 labscript = sequence_df['labscript'].iloc[0]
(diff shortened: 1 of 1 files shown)


lakodarmstadt:
Somehow my PR did not pop up in my mail notifications. Maybe this is because I generated it with gitkraken.
Furthermore I was wondering if some possibility to load depricated msgpack-DF is desirable. One option could be: https://pandas-msgpack.readthedocs.io/en/latest/install.html

dihm:
As discussed in #82, going to use pickle instead of parquet format. Declining this in favor of #95.

PR #93, opened 2021-07-12
Update lyse docs

dihm:
Updating the lyse docs to follow the same paradigm as all the other modules.

It now uses recursive autosummary calls to generate documentation stubs for the module components. This also increases coverage to the GUI code as well.


dihm:
Ready to go. Docstring coverage is 15.2%.

PR #97, opened 2022-02-16
Lyse simple data

ispielma:
Updated the Run object's get_trace method to have a "raw_data" kwarg.  This instructs lyse to simply return the data without interpreting it as a time trace.  This solves an issue with the TeX scope class which saves data but not time information.

### DIFF
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -388,11 +388,15 @@ class Run(object):
                 raise Exception('The group \'%s\' does not exist'%group)
             return get_attributes(h5_file[group])
 
-    def get_trace(self,name):
+    def get_trace(self, name, raw_data=False):
         """Return the saved data trace `name`.
         
         Args:
             name (str): Name of saved data trace to get.
+            
+        Kwargs:
+            raw_data (bool): option to return the h5_data directly
+            without interperting it as a 2-D time trace.
 
         Raises:
             Exception: If `name` trace does not exist.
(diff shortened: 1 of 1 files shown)


lyse/__init__.py
```diff
@@ -388,11 +388,15 @@ def get_attrs(self, group):
...
-    def get_trace(self,name):
+    def get_trace(self, name, raw_data=False):
         """Return the saved data trace `name`.
         
         Args:
             name (str): Name of saved data trace to get.
+            
+        Kwargs:
+            raw_data (bool): option to return the h5_data directly
+            without interperting it as a 2-D time trace.
```
dihm, line 399: While this would be great syntax, google style docstrings don't make a distinction between args and kwargs. Could you instead put this as:
```
Args:
    name (str): Name of saved data trace to get.
    raw_data (bool, optional): option to return the h5_data directly 
        without interpreting it as a 2-D time trace.
```

lyse/__init__.py
```diff
@@ -405,7 +409,13 @@ def get_trace(self,name):
             if not name in h5_file['data']['traces']:
                 raise Exception('The trace \'%s\' does not exist'%name)
             trace = h5_file['data']['traces'][name]
-            return array(trace['t'],dtype=float),array(trace['values'],dtype=float)         
+            
+            if raw_data:
+                data = trace[:]
```
dihm, line 414: Could we change this to the more general (but less intuitive looking) `data = trace[()]`?

dihm (changes requested):
Looks good to me. I'm all for extending the flexibility of lyse in ways people actually want to use it. I just request a couple minor tweaks then I'll go ahead and merge.

ispielma:
The pull request has been modified with these changes.

lyse/__init__.py
ispielma: The pull request has been modified with these changes.

PR #98, opened 2022-05-09
Definition of new function "get_waits"

th0mam:
I defined a new function "get_waits".  Returns a 2D array of times and timeout statuses.
This may be quite useful to some teams, where timeout signifies faulty experiment or loss of the atom from the trap.

### DIFF
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -407,6 +407,22 @@ class Run(object):
             trace = h5_file['data']['traces'][name]
             return array(trace['t'],dtype=float),array(trace['values'],dtype=float)         
 
+    def get_waits(self):
+        """Return the time out status of waits.
+
+        Raises:
+            Exception: If the experiment has no waits.
+
+        Returns:
+            :obj:`numpy:numpy.ndarray`: Returns 2-D timetrace of times `'time'`
+            and time out status `'timed_out'`.
+        """
+        with h5py.File(self.h5_path,'r') as h5_file:
+            if not 'waits' in h5_file['data']:
+                raise Exception('The shot has no waits')
+            trace=h5_file['data']['waits']
+            return array((trace['time'],trace['timed_out']),dtype=float)  
+        
     def get_result_array(self,group,name):
         """Returns saved results data.
 


lyse/__init__.py
```diff
@@ -407,6 +407,22 @@ def get_trace(self,name):
...
+            Exception: If the experiment has no waits.
+
+        Returns:
+            :obj:`numpy:numpy.ndarray`: Returns 2-D timetrace of times `'time'`
+            and time out status `'timed_out'`.
+        """
+        with h5py.File(self.h5_path,'r') as h5_file:
+            if not 'waits' in h5_file['data']:
+                raise Exception('The shot has no waits')
+            trace=h5_file['data']['waits']
```
dihm, line 423: could we rename this variable from `trace` to `waits`, so it is more descriptive?

lyse/__init__.py
```diff
@@ -407,6 +407,22 @@ def get_trace(self,name):
...
+            return array((trace['time'],trace['timed_out']),dtype=float)
```
dihm, line 424: The `waits` object actually includes quite a bit more useful information than just the script start time and timeout state. Would you be opposed to returning all of the information (ie wait label, script start time, timeout time, actual length of the wait, and timeout state)?

Admittedly, that object is a [structured numpy array](https://numpy.org/doc/stable/user/basics.rec.html) where each column has a unique label and dtype, making the use a little different than normal numpy arrays. To get access to things, you'd need to use the label indexing.

Examples:
```python
waits = h5_file['data']['waits'][()]  # last bit is necessary to convert from h5py dataset to a numpy array
return waits
```
```python
waits.dtype  # [('label', 'S256'), ('time', '<f8'), ('timeout', '<f8'), ('duration', '<f8'), ('timed_out', '?')]
waits[0]  # gives the first wait object, note that printing this looks like a normal tuple
waits['timed_out']  # gives a boolean numpy array of the timed_out column
waits['label']  # gives a bytestring numpy array of the labels
```

lyse/__init__.py
```diff
@@ -407,6 +407,22 @@ def get_trace(self,name):
...
+
+        Raises:
+            Exception: If the experiment has no waits.
+
+        Returns:
+            :obj:`numpy:numpy.ndarray`: Returns 2-D timetrace of times `'time'`
+            and time out status `'timed_out'`.
+        """
+        with h5py.File(self.h5_path,'r') as h5_file:
+            if not 'waits' in h5_file['data']:
```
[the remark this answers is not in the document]
th0mam: I think I can add the extra check.
Out of curiosity I tested the error today with the function get_trace using arbitrary trace name without the data subgroup present and I got the same noisy error. 
I also tested the get_traces function, which simply returns empty array.

lyse/__init__.py
th0mam: Great suggestion, in the end I used much simpler solution returning all of the parameters in structured array and omitted the variable all together

lyse/__init__.py
th0mam: For my application I didn´t need other parameters, but it is no issue to return all the information.

th0mam:
I believe I resolved the issue with the noisy error. I also added the requested versatility to the function, returning all of the information available, and the function get_wait returning the parameters of a wait by name.
The function get_wait returns a tuple. I didn't manage to get the structured dataset, but I believe it is not necessary, as the object is quite simple to work with.

lyse/__init__.py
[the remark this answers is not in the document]
dihm: Yeah. I also noticed that about get_trace when looking this over. I appreciate you adding the check here though. With time I imagine a similar check will make its way there as well.

dihm (approved):
I'm going to go ahead an approve this, but it will take me a couple days to merge so I can find the time to test it out first.

Again, thanks for all your work. It is very much appreciated.

dihm:
Actually, got to it faster than I thought. Looks good, I'll go ahead and merge now.

PR #102, opened 2023-03-11
Implement optimization described in #92

### LINKED ISSUE #92: lyse speedup and the analysis plot panel
Hey everyone,

For our experiment at EPFL, we were writing a plotting library [analysis_plot_panel](https://lqg-epfl.github.io/analysis_plot_panel/)  that relies on the pyqtgraph. Our experimental rate is higher than 1 Hz. This means that we had to streamline our analysis scripts. The issue was that Matplolib takes a significant time (sometimes more than 1s) to update simple plots. This meant that we had to deactivate the plots in some cases. This is of course not optimal when one whats to debug an experiment.

The analysis_plot_panel solves this issue since pyqtgraph allows for high-speed refreshes (>100 Hz) for 1d and 2d figures. The script is integrated into lyse using a single shot routine. Check the [documentation](https://lqg-epfl.github.io/analysis_plot_panel/) for details.

The reason we are sharing this here as an issue is, that during our research to find out how to speedup lyse, we learned that a lot of the time of the analysis is wasted at the loading and saving stage (from h5_file). The reason for this was not the data transfer, but the fact, that at every save_result call the h5 file would be opened and closed again. This takes significantly more time than saving the actual result. We solved this issue by rewriting the functions in the __init__.py file. Now they take h5_file as an argument. This means that if you call save_results it opens the file only once and hands the file to all of the save_result calls. Would there be a possibility to integrate this feature into lyse? The new __init__.py file can be found [here](https://github.com/LQG-EPFL/analysis_plot_panel/blob/master/new_lyse_init/__init__.py).

dihm:
Implements a wrapper function that allows all lyse functions to optionally take a handle to the h5_file instead of opening and closing for each operation. This allows a single open and close operation for a single shot run.

This tends to speed up writes more than reads.

This closes #92

### DIFF
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -67,7 +67,7 @@ Single shot analysis with global file opening
 	with Run(path).open('r+') as shot:
 
 		# Obtaining a trace:
-		t, mot_fluorecence = run.get_trace('mot fluorecence')
+		t, mot_fluorecence = shot.get_trace('mot fluorecence')
 
 		# Now we might do some analysis on this data. Say we've written a
 		# linear fit function (or we're calling some other libaries linear
--- a/docs/source/examples.rst
+++ b/docs/source/examples.rst
@@ -53,6 +53,38 @@ An analysis on a single shot
 	# shots in a multishot analysis:
 	run.save_result('mot loadrate', c)
 
+Single shot analysis with global file opening
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: python
+
+	from lyse import *
+
+	# Instantiate Run object and open
+	# Globally opening the shot keeps the h5 file open
+	# This prevents excessive opening and closing of the file
+	# which can slow down the analysis
+	with Run(path).open('r+') as shot:
+
+		# Obtaining a trace:
+		t, mot_fluorecence = run.get_trace('mot fluorecence')
+
+		# Now we might do some analysis on this data. Say we've written a
+		# linear fit function (or we're calling some other libaries linear
+		# fit function):
+		m, c = linear_fit(t, mot_fluorecence)
+		int_tot = mot_fluorecence.sum()
+		mf_min = mot_fluorecence.min()
+		mf_max = mot_fluorecence.max()
+
+		normalised_fluorecence = normalise_response(mot_fluorecence)
+
+		# We might wish to save this result so that we can compare it across
+		# shots in a multishot analysis:
+		shot.save_result('mot loadrate', c)
+		shot.save_result('mot integrated', int_tot)
+		shot.save_results('mot fl min', mf_min, 'mot fl max', mf_max)
+		shot.save_result_array('norm mot fluorecence', normalised_fluorecence)
 
 An analysis on multiple shots
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -22,6 +22,7 @@ import inspect
 import sys
 import threading
 import functools
+import contextlib
 
 import labscript_utils.h5_lock, h5py
 from labscript_utils.labconfig import LabConfig
@@ -241,6 +242,7 @@ class Run(object):
     def __init__(self,h5_path,no_write=False):
         self.__h5_path = h5_path
         self.__no_write = no_write
+        self.__h5_file = None
         self.__group = None
         if not self.no_write:
             self._create_group_if_not_exists(h5_path, '/', 'results')
@@ -287,12 +289,76 @@ class Run(object):
     def _path_to_file_name(self, path):
         file_name = os.path.basename(path).split('.py')[0]
         return file_name
+
+    @property
+    def h5_path(self):
+        """str: The value provided for `h5_path` during instantiation."""
+        return self.__h5_path
+
+    @property
+    def no_write(self):
+        """bool: The value provided for `no_write` during instantiation."""
+        return self.__no_write
+    
+    @property
+    def h5_file(self):
+        """h5py.File: opened h5py file handle for the shot"""
+        return self.__h5_file
+    
+    @contextlib.contextmanager
+    def open(self, mode):
+        """Opens the Run's h5 file for successive reads/writes.
+
+        Intended to be used as a context manager.
+        Supports all h5 modes, though only `'r'` and `'r+'`/`'a'` are typically
+        used within lyse itself.
+
+        Args:
+            mode (str): which mode to open the h5 file with.
+                Must be 'r', 'a', 'r+', 'w', 'w-', or 'x'.
+                Lyse typically only uses 'r' and 'a'.
+
+        Examples:
+            >>> from lyse import *
+            >>> shot = Run(path)
+            >>> with shot.open('a'):
+            >>>     # shot processing that requires reads/writes
+            >>>     t, vals = shot.get_trace('my_trace')
+            >>>     results = vals**2
+            >>>     shot.save_result_array('my_result', results)
+            >>> # Shot processing that doesn't require h5 reads/writes
+            >>> print(f'Max of results is {results.max():.3f}')
+            Max of results is 5.003
+
+        Raises:
+            PermissionError: If the Run is set as read-only but a write mode is requested
+        """
+
+        # ensure no_write is respected
+        if self.no_write:
+            if mode == 'r':
+                pass
+            else:
+                msg = f'Cannot open file with a write mode; this run is read-only'
+                raise PermissionError(msg)
+            
+        # ensure file is not already opened
+        if self.__h5_file is not None:
+            msg = 'h5_file already opened'
+            raise PermissionError(msg)
+        
+        self.__h5_file = h5py.File(self.h5_path, mode)
+        try:
+            yield None
+        finally:
+            self.__h5_file.close()
+            self.__h5_file = None
             
     def open_file(mode):
-        """Wrapper for lyse functions to allow passing in an optional open h5file instance
+        """Wrapper for lyse functions to allow using previously opened file with context manager.
         
         If multiple read/write operations happen on a Run in a single shot,
-        opening the h5file once and passing it to the necessary functions
+        opening the h5file once via the context manager `open`
         can speed up the analysis execution time by limiting the number of times
         the file must be opened/closed.
 
@@ -306,39 +372,29 @@ class Run(object):
         """
         def decorator_open_file(func):
             @functools.wraps(func)
-            def wrapper(self, *args, h5_file=None, **kwargs):
+            def wrapper(self, *args, **kwargs):
 
                 if self.no_write:
                     if mode == 'r':
                         pass
-                    elif mode in ['a', 'r+', 'w', 'w-', 'x']:
+                    else:
                         msg = f'Cannot perform operation {func.__name__:s}; this run is read-only'
                         raise PermissionError(msg)
 
-                if h5_file is None:
-                    with h5py.File(self.h5_path, mode) as h5_file:
-                        return func(self, *args, h5_file = h5_file, **kwargs)
+                if self.__h5_file is None:
+                    with self.open(mode):
+                        return func(self, *args, **kwargs)
                 else:
-                    if (h5_file.mode == 'r') and (mode != 'r'):
+                    if (self.__h5_file.mode == 'r') and (mode != 'r'):
                         msg = (f"Cannot perform operation {func.__name__:s}; "
                             + f"requested mode {mode:s} not compatible with "
-                            + f"h5_file's mode {h5_file.mode:s}")
+                            + f"h5_file's mode {self.__h5_file.mode:s}")
                         raise PermissionError(msg)
                     else:
-                        return func(self, *args, h5_file = h5_file, **kwargs)
+                        return func(self, *args, **kwargs)
                 
             return wrapper
         return decorator_open_file
-
-    @property
-    def h5_path(self):
-        """str: The value provided for `h5_path` during instantiation."""
-        return self.__h5_path
-
-    @property
-    def no_write(self):
-        """bool: The value provided for `no_write` during instantiation."""
-        return self.__no_write
 
     @property
     def group(self):
@@ -400,13 +456,8 @@ class Run(object):
         self.__group = groupname
 
     @open_file('r')
-    def trace_names(self, h5_file=None):
+    def trace_names(self):
         """Return a list of all saved data traces in Run.
-
-        Args:
-            h5_file (h5py.File, optional): Allows passing in a handle
-                to an already opened h5 file. If None, will open in
-                read only mode and close after operation.
 
         Raises:
             KeyError: If the group `'/data/traces/'` does not yet exist.
@@ -415,19 +466,16 @@ class Run(object):
             list: List of keys in the h5 file's `'/data/traces/'` group.
         """
         try:
-            return list(h5_file['data']['traces'].keys())
+            return list(self.h5_file['data']['traces'].keys())
         except KeyError:
             return []
 
     @open_file('r')                
-    def get_attrs(self, group, h5_file=None):
+    def get_attrs(self, group):
         """Returns all attributes of the specified group as a dictionary.
 
         Args:
             group (str): Group for which attributes are desired.
-            h5_file (h5py.File, optional): Allows passing in a handle
-                to an already opened h5 file. If None, will open in
-                read only mode and close after operation.
 
         Raises:
             Exception: If the `group` does not exist.
@@ -435,21 +483,18 @@ class Run(object):
         Returns:
             dict: Dictionary of attributes.
         """
-        if not group in h5_file:
+        if not group in self.h5_file:
             raise Exception('The group \'%s\' does not exist'%group)
-        return get_attributes(h5_file[group])
+        return get_attributes(self.h5_file[group])
 
     @open_file('r')
-    def get_trace(self, name, raw_data=False, h5_file=None):
+    def get_trace(self, name, raw_data=False):
         """Return the saved data trace `name`.
         
         Args:
             name (str): Name of saved data trace to get.
             raw_data (bool, optional): option to return the h5_data directly 
                 without interpreting it as a 2-D time trace.
-            h5_file (h5py.File, optional): Allows passing in a handle
-                to an already opened h5 file. If None, will open in
-                read only mode and close after operation.
 
         Raises:
             Exception: If `name` trace does not exist.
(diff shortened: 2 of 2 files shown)


dihm:
Using a very simple test analysis script with crude timing seems to suggest that reads get little to modest improvements (up to x1.2, potentially being elided by numpy magic), and writes get a more significant improvement (x20-x40).

```python
from lyse import *
import time

run = Run(path)

start = time.time()
for _ in range(10):
    _, _ = run.get_trace('ClockTest')

end = time.time()
orig_read = end-start
print(f'Elapsed original read: {orig_read:.3e} s')

start = time.time()
with h5py.File(run.h5_path, 'r') as f:
    
    for _ in range(10):
        _, _ = run.get_trace('ClockTest', h5_file=f)

end = time.time()

new_read = end - start
print(f'Elapsed new read: {new_read:.3e} s')
print(f'New is {orig_read/new_read:.3f} times faster')


start = time.time()
for _ in range(100):
    run.save_result('ClockTest',5)

end = time.time()
orig_write = end-start
print(f'Elapsed original write: {orig_write:.3e} s')

start = time.time()
with h5py.File(run.h5_path, 'r+') as f:

    for _ in range(100):
       run.save_result('ClockTest', 5, h5_file=f)

end = time.time()
new_write = end-start
print(f'Elapsed new write: {new_write:.3e} s')
print(f'New is {orig_write/new_write:.3f} times faster')
```

A typical output from this script is
```shell
Elapsed original read: 8.700e-01 s
Elapsed new read: 8.535e-01 s
New is 1.019 times faster
Elapsed original write: 7.203e-01 s
Elapsed new write: 1.997e-02 s
New is 36.076 times faster
```

dihm:
Assuming there aren't objections, I'd like to add something to the documentation describing how to do this more clearly before merging.

philipstarkey:
I think a better way to do this might be to add method to the run class that is decorated with `contextmanager` that opened the file. Then you could store the open file on the run instance, detect it's open, and not have to pass it in. Should make adapting existing analysis code easier as method signature would change, you just need to add a `with run.open('w'):` or something like that to your analysis code.

dihm:
That makes sense. After editing three method signatures I had a quick moment of "there must be a better way" then promptly moved on just so I could have something to test that the speed-ups are actually worth it. I'll give it a better crack this time and hopefully the result will be a bit more user friendly.

dihm:
I've had another pass at it. One nice benefit to this method is that existing methods which make successive calls to the lyse API (ie `save_result_arrays`) now automatically benefit from this without any change on the user's part.

A potential downside is that mixing methods using the contextmanager with the older per-method implementation (eg local user modifications to extend lyse functionality with new Run.methods) is all but certain to lead to multiple attempts to open the same file. Users will be forced to update any local changes to lyse to match this style if they want to take advantage of the new feature.

dihm:
@philipstarkey If you have another moment, I'd appreciate your eyes on this updated PR before I merge it.

Maybe my one question is whether the `open_file` wrapper is implemented correctly, particularly if it should actually be a method of the `Run` class?

philipstarkey:
I'll try and have a look this weekend!

lyse/__init__.py
```diff
@@ -296,6 +299,102 @@ def h5_path(self):
...
+        if self.__h5_file is not None:
+            msg = 'h5_file already opened'
+            raise PermissionError(msg)
+        
+        self.__h5_file = h5py.File(self.h5_path, mode)
+        try:
+            yield None
+        finally:
+            self.__h5_file.close()
+            self.__h5_file = None
```
philipstarkey, line 355: This should probably also use a context manager. Something like:

```
with h5py.File(self.h5_path, mode) as f:
    self.__h5_file = f
    try:
        yield f
    finally:
        self.__h5_file = None
```

That way you both yield the file object like h5py.File does, and you leave it up the the h5py context manager to close the file properly.

lyse/__init__.py
```diff
@@ -296,6 +299,102 @@ def h5_path(self):
...
+            PermissionError: If the Run is set as read-only but a write mode is requested
+        """
+
+        # ensure no_write is respected
+        if self.no_write:
+            if mode == 'r':
+                pass
+            else:
+                msg = f'Cannot open file with a write mode; this run is read-only'
+                raise PermissionError(msg)
```
philipstarkey, line 343: This could be condensed to `if self.no_write and mode != “r”:`

lyse/__init__.py
```diff
@@ -296,6 +299,102 @@ def h5_path(self):
...
+            
+        # ensure file is not already opened
+        if self.__h5_file is not None:
+            msg = 'h5_file already opened'
+            raise PermissionError(msg)
```
philipstarkey, line 348: If we new the mode wasn’t changing, there is nothing stopping us yielding `self.__h5_file` here (and not closing the file afterwards). Does that go some way to helping with resuable code snippets that might not know if they’re being called inside the context manager already?

lyse/__init__.py
```diff
@@ -296,6 +299,102 @@ def h5_path(self):
...
+        
+        self.__h5_file = h5py.File(self.h5_path, mode)
+        try:
+            yield None
+        finally:
+            self.__h5_file.close()
+            self.__h5_file = None
+            
+    def open_file(mode):
```
philipstarkey, line 357: I think this decorator doesn’t need to be inside the `Run` class? Probably clearer if it’s not (otherwise it’s odd that it’s not defined as a `staticmethod` and also doesn’t take `self` as an argument)

lyse/__init__.py
```diff
@@ -296,6 +299,102 @@ def h5_path(self):
...
+        def decorator_open_file(func):
+            @functools.wraps(func)
+            def wrapper(self, *args, **kwargs):
+
+                if self.no_write:
+                    if mode == 'r':
+                        pass
+                    else:
+                        msg = f'Cannot perform operation {func.__name__:s}; this run is read-only'
+                        raise PermissionError(msg)
```
philipstarkey, line 382: If statement can also be condensed as above

lyse/__init__.py
```diff
@@ -296,6 +299,102 @@ def h5_path(self):
...
+
+                if self.__h5_file is None:
+                    with self.open(mode):
+                        return func(self, *args, **kwargs)
+                else:
+                    if (self.__h5_file.mode == 'r') and (mode != 'r'):
+                        msg = (f"Cannot perform operation {func.__name__:s}; "
+                            + f"requested mode {mode:s} not compatible with "
+                            + f"h5_file's mode {self.__h5_file.mode:s}")
+                        raise PermissionError(msg)
```
philipstarkey, line 392: I think some additional documentation around this (either comment here or in docstring) would be useful. It took me a while to work out that passing in `r+` as the mode (e.g. see use of decorator further down file) wasn’t incompatible with someone manually opening the file with `a` for example. I think this is something that will be easily forgotten unless we document it now. The only incompatible case is the one for this error message, where if the file is open in read only, then any decorator use that wants write access needs to throw an error.

That could be an argument for not actually having the argument to the decorator be `mode`, but a `Boolean` that indicates whether we need it to be open in a writable mode or not. That boolean would then toggle between `r` or `a` in the call to `self.open()` a few lines above.

philipstarkey: Which brings me to another point, it seems like we’ve switched the use of `a` to `r+` which would be a backwards incompatible change?

lyse/__init__.py
```diff
@@ -553,31 +654,27 @@ def save_result(self, name, value, group=None, overwrite=True):
...
+        if not group:
+            if self.group is None:
+                msg = """Cannot save result; no default group set. Either
+                    specify a value for this method's optional group
+                    argument, or set a default value using the set_group()
+                    method."""
+                raise ValueError(dedent(msg))
+            # Save to analysis results group by default
+            group = 'results/' + self.group
+        elif not group in self.h5_file:
```
philipstarkey, line 666: Unrelated to this PR, but can you change this to `group not in self.h5_file` while you are here? It is equivalent, but more readable syntax

philipstarkey: Looks like there are a couple of others too in some later methods

philipstarkey (changes requested):
Looking good. Got some suggested changes. Hopefully not too nitpicky…I realised afterwards that I reviewed this like I would something at work 😅

lyse/__init__.py
dihm: It's context managers all the way down!

lyse/__init__.py
dihm: Well noted! This is the result of simplifying the even more convoluted logic tree I started with.

lyse/__init__.py
dihm: I thought that would be the case, just wanted to get another opinion.

Now that I look at it, I'm pleasantly surprised it even works as is. It's not like I called it as a typical class method. TIL more about what using self actually does.

lyse/__init__.py
dihm: Fair enough. Just need to make sure the modes are compatible, then return.

dihm: Actually, I don't think we have to yield the file handle (normal operation doesn't). If fact, I think it has to `return None` here to match signatures and prevent executing the file open, right?

lyse/__init__.py
dihm: Yeah, this is due to the quirks of how h5py File objects work. While there are a number of mode options for [opening the file](https://docs.h5py.org/en/stable/high/file.html#opening-creating-files), the file itself can only be in [two modes](https://docs.h5py.org/en/stable/high/file.html#h5py.File.mode) (`'r'` and `'r+'`). All the other options for opening the file deal exclusively with how to handle creating a new file. I'll update the docstring to make this clear as it isn't super intuitive at first glance.

This relates to my thought process for changing most of the calls from `a` to `r+`. The only difference between these two is that `a` will create the file if it doesn't exist already. It is my understanding that almost all lyse functions should never create a new file on their own: that job is handled by runmanager or the `Sequence` constructor. For example, if `save_result` is creating a new file, I'd argue that is a bug that should raise an error to be fixed instead of failing silently as it would now. Assuming that reasoning is valid, I see there are a few docstrings that should be updated to reflect that. I've gone ahead and changed them. If my reasoning is wrong, I can change things back.

I decided to keep mode strings (instead of a bool) to match the h5py documentation. Felt like it would be more clear what was happening if we didn't add another layer of abstraction over top.

lyse/__init__.py
dihm: I think I have found them all. There were quite a few.

lyse/__init__.py
dihm: Ah, should have tested this out before pushing. Naive putting the decorator outside the class [doesn't work](https://stackoverflow.com/questions/1263451/python-decorators-in-classes) because you need to access the class instance from within the wrapper and that doesn't work unless we define within the class. Not sure how much I trust a single SO post for best practices, but what I had appears to be a fairly common pattern. I'll revert for now.

dihm:
Don't worry about nitpicking. I'm a serious nitpicker on reviews myself 😉 I appreciate the effort it takes to put extra eyes on someone else's code.

lyse/__init__.py
philipstarkey: This really should have worked. `self` isn't accessed until the method is called (e.g. it's used inside `wraps` which is the replacement method returned by the decorator). It's equivalent to the `define_state` decorator in BLACS. The only difference is the use of the `functools.wraps`. that should be fine butmaybe I'm missing something. Is the use of wraps meant to be on the middle function instead? It's primary purpose is to copy the docstring of the `Run` methods decorated with `open`. Should be easy enough to reason out where it needs to be and test it's copying the doc strings

lyse/__init__.py
philipstarkey: A reason for `a` might be if you're using `Run` as a multishot analysis storage system. I suppose that's probably a niche use case and we can just require you create the file first in that case.

Thoughts @chrisjbillington ?

lyse/__init__.py
philipstarkey: You need to prevent the later open and yield occurring. I think you still want to use yield too though. Maybe:

```
    yield self.__h5_file
    return
```

Otherwise you don't get the file handle out the second time, so you get incompatible behaviour of you reuse some code inside an already open context manager. And I don't know how the `contextmanager` decorator handles no yield.

lyse/__init__.py
chrisjbillington: It is niche, and we shouldn't worry much about niche things.

Though for what it's worth, I also think worrying about someone passing in a non-existent file unintentionally is fairly niche. So I don't think it really matters.

On balance though, I lean toward discouraging using `Run` for multi-shot storage, because that means we have to continue to commit to ensuring it works with arbitrary HDF5 files, whereas it would be nice to be able to assume the HDF5 file is an actual shot file. It currently works with arbitrary files, but minimising future commitments is good.

And if someone wants to use HDF5 to store data, they may do so with `h5py` directly - our wrapper is not adding much and is actually pretty limiting. I would encourage using raw `h5py` instead for this, so I have no problem making `Run` less convenient to use for this, it is unlikely to be the best tool for the job anyway.

So +1 for `r+` I think.

lyse/__init__.py
dihm: Well I can report that contextManager doesn't like it if there isn't a yield since it is a generator. TIL.

I'm still confused why it should yield the file handle instead of `None`. The file handle isn't returned by this context normally; it just saves it as a Run attribute.

I suppose that means we can't do `with run.open('r') as f`, but my working assumption is that we didn't want people to normally access the file directly. If they really need the handle, they can get it from the `run.h5_file` property. I think I slightly prefer that (as it discourages casual direct interaction with the h5 file since you need to separately get the handle) but I don't have strong feelings about it. Either is fine.

lyse/__init__.py
dihm: Cool. I'll keep as-is then.

Also, I'm pretty sure `Run` will currently fail if called on a non-existing file thanks to `_create_group_if_not_exists`, which already uses `r+`, so hopefully this won't break anyone's workflow.

https://github.com/labscript-suite/lyse/blob/82db03237f8667ba65859c1fb9e647a123bfb98e/lyse/__init__.py#L331-L341

lyse/__init__.py
philipstarkey: The reason to yield the file handle is so it mimics `h5py.File(..)`. Probably most people won't use it. And yep, you could access it via the attribute on the `Run` instance. But given we have to yield, we may as well yield something useful! You don't have to even assign it (`with run.open(mode)` should work as well as `with run.open(mode) as f`)

philipstarkey: Think I must have not read the whole comment before I replied. Feel like I've just repeated what you said. Oops. I don't see any harm in given people the file handle though. Sometimes you might want direct access to the file. Generally you expect context managers that open a file to yield the file handle as well. I think my feelings are mainly driven by that - trying to be consistent with how you'd expect it to behave.

lyse/__init__.py
philipstarkey: Oh, I see. Given you don't need to write to that attribute inside the decorator, can you just switch it to using the `self.h5_file` property instead? Then the decorator could live outside of the class.

dihm: OK, latest push should resolve this one too.

dihm:
I think everything is addressed. Ready for a re-review at your leisure.

lyse/__init__.py
chrisjbillington: For what it's worth, the name mangling is achieving nothing and could be removed. The purpose of double-underscore mangling is to make attributes inaccessible from subclasses or parent classes, which we are not trying to do.

There's not even much reason to make them single-underscore private, Id' argue since they all have getters (I guess some don't have setters). My guess is that the main reason they are private at all is to make it easier to document them by adding docstrings to the getters.

But yeah, single-underscore is arguable, whereas double is achieving nothing.

philipstarkey: The name mangling, and exposure via property getters, is to prevent end users from changing the internal state. A lot of the checks, etc. rely on the assumption that these variables are private. If one of those mangled variables was changed (accidentally) I think you'd see very odd behaviour.

chrisjbillington: Well that's an argument for them being private generally (i.e. one leading underscore), which I lean against more than you but accept there's a case for. Name mangling as triggered by two leading underscores is for privacy from methods of subclasses and parent classes though, which I don't think there is a case for.

philipstarkey:
I'm happy with this.

One thought I did have was whether it made more sense to `yield self` rather than the file handle. It's more useful, because you can then do `with Run(path).open("r+") as shot:`...not sure if it's more confusing though (not that you'd be forced to use it)

dihm:
> One thought I did have was whether it made more sense to yield self rather than the file handle. It's more useful, because you can then do with Run(path).open("r+") as shot:...not sure if it's more confusing though (not that you'd be forced to use it)

I like the idea of this. It helps hide the h5 handle and makes for a decent whole-shot drop in replacement line. It does lend itself to strangeness like `run= Run(path); with run.open('r+') as shot: ...`, but some decent examples may help prevent that from happening.

 I'll go ahead and implement it as I think it is better.

docs/source/examples.rst
```diff
@@ -53,6 +53,38 @@ An analysis on a single shot
...
+	from lyse import *
+
+	# Instantiate Run object and open
+	# Globally opening the shot keeps the h5 file open
+	# This prevents excessive opening and closing of the file
+	# which can slow down the analysis
+	with Run(path).open('r+') as shot:
+
+		# Obtaining a trace:
+		t, mot_fluorecence = run.get_trace('mot fluorecence')
```
philipstarkey, line 70: Needs to be `shot.`

lyse/__init__.py
```diff
@@ -365,22 +365,45 @@ def open(self, mode):
...
 
-        Raises:
-            PermissionError: If the Run is set as read-only but a write mode is requested
+            Open and create the shot handle for the whole analysis
+            in a single line.
+
+            >>> from lyse import *
+            >>> with Run(path).open('r+') as shot:
+            >>>     # shot processing that requires reads/writes
+            >>>     t, vals = shots.get_trace('my_trace')
```
philipstarkey, line 399: Should be `shot.`

lyse/__init__.py
```diff
@@ -365,22 +365,45 @@ def open(self, mode):
...
+            these two openings into one.
+
             >>> from lyse import *
             >>> shot = Run(path)
-            >>> with shot.open('r+'):
+            >>> with shot.open('r'):
             >>>     # shot processing that requires reads/writes
-            >>>     t, vals = shot.get_trace('my_trace')
+            >>>     _, vals = shot.get_trace('my_trace')
+            >>> with shot.open('r+')
```
philipstarkey, line 384: Like should end with a `:`?

philipstarkey (changes requested):

docs/source/examples.rst
dihm: This is what I get for typing and not checking...

dihm:
Dumb typos fixed. Thanks for catching those @philipstarkey!

philipstarkey (approved):

PR #103, opened 2023-03-13
Fix signature mis-match when loading custom FigureManager backend

### LINKED ISSUE #101: Bug: matplotlib.pyplot error
## Description
Hi *labscript* Team!

For some analysis scripts to plot data, I get an error from pyplot. This happens only the first time running the script after launching Lyse, and only if matplotlib has to call figure(), subplots() or similar (just using  pyplot.plot does not cause the error for some reason).

I have matplotlib version 3.5.3 and lyse  3.1.0rc2.dev84+g1f0c876.

## Example

The error can be reproduced by running the following script on the first shot after launching Lyse:
```
import matplotlib.pyplot as plt

fig, ax = plt.subplots(2,1)

# further code for plotting...
```

This is the full error traceback that is printed:
```
File "/home/johannes/labscript-suite/lyse/lyse/analysis_subprocess.py", line 323, in do_analysis
    exec(code, self.routine_module.__dict__)

  File "/home/johannes/labscript-suite/userlib/analysislib/leolab/test.py", line 5, in <module>
    fig, ax = plt.subplots(2,1)

  File "/home/johannes/.conda/envs/labscript/lib/python3.9/site-packages/matplotlib/pyplot.py", line 1453, in subplots
    fig = figure(**fig_kw)

  File "/home/johannes/labscript-suite/lyse/lyse/figure_manager.py", line 77, in __call__
    number, fig =  self.get_first_empty_figure(identifier, *args,**kwargs)

  File "/home/johannes/labscript-suite/lyse/lyse/figure_manager.py", line 33, in get_first_empty_figure
    fig = self._figure(i,*args,**kwargs)

  File "/home/johannes/.conda/envs/labscript/lib/python3.9/site-packages/matplotlib/pyplot.py", line 806, in figure
    manager = new_figure_manager(

  File "/home/johannes/.conda/envs/labscript/lib/python3.9/site-packages/matplotlib/pyplot.py", line 324, in new_figure_manager
    _warn_if_gui_out_of_main_thread()

  File "/home/johannes/.conda/envs/labscript/lib/python3.9/site-packages/matplotlib/pyplot.py", line 314, in _warn_if_gui_out_of_main_thread
    if (_get_required_interactive_framework(_get_backend_mod())

  File "/home/johannes/.conda/envs/labscript/lib/python3.9/site-packages/matplotlib/pyplot.py", line 217, in _get_backend_mod
    switch_backend(dict.__getitem__(rcParams, "backend"))

  File "/home/johannes/.conda/envs/labscript/lib/python3.9/site-packages/matplotlib/pyplot.py", line 301, in switch_backend
    globals()[func_name].__signature__ = inspect.signature(

AttributeError: 'method' object has no attribute '__signature__'
```

## Explanation and possible solution

The error above in the line `globals()[func_name].__signature__ = inspect.signature(` happens for `func_name="show"`, when matplotlib does something for registering the backend. This function in the globals was overwritten in [`figure_manager.py`](https://github.com/labscript-suite/lyse/blob/master/lyse/figure_manager.py#L143) from the default `show` function of pyplot. 

Now it turns out that here it makes a difference if `show` is defined as a function, as in pyplot, or as a method of the class FigureManger in Lyse, because for the method we cannot set the `__signature__` how to call is as attribute. 
A quite simple solution to resolve the error is to make sure `show` is not a class method but a function by changing the linked line in `figure_manager.py` to:

`matplotlib.pyplot.show = lambda: figuremanager.show`

Do you think that changing this line makes sense? Or should I create subplots differently in lyse?
I did not yet really understand why we need to overwrite some of the pyplot functions.

dihm:
Fixes #101

### DIFF
--- a/lyse/figure_manager.py
+++ b/lyse/figure_manager.py
@@ -140,4 +140,4 @@ def install():
     figuremanager = FigureManager()
     matplotlib.pyplot.figure = figuremanager
     matplotlib.pyplot.close = figuremanager.close
-    matplotlib.pyplot.show = figuremanager.show
+    matplotlib.pyplot.show = lambda: figuremanager.show


lyse/figure_manager.py
```diff
@@ -140,4 +140,4 @@ def install():
     figuremanager = FigureManager()
     matplotlib.pyplot.figure = figuremanager
     matplotlib.pyplot.close = figuremanager.close
-    matplotlib.pyplot.show = figuremanager.show
+    matplotlib.pyplot.show = lambda: figuremanager.show
```
philipstarkey, line 143: @dihm Doesn’t this need to be `lambda: figuremanager.show()`?

dihm: Interesting. Apparently both ways work. I'll admit my understanding of `lambda` isn't very deep, but now that you point it out I suspect you are correct in the strictest sense. Does it work as-is because it's just passing through a passed-through function handle?

I'm not sure it is worth changing, but I will if you think it is.

chrisjbillington: Definitely needs to be `lambda: figuremanager.show()`.

`lambda: figuremanager.show` does nothing, it defines a function that returns a function without calling it.  It appears to 'work' because explicitly calling `show()` is not necessary when running an analysis routine within lyse - all plots got shown regardless.

But when running an analysis routine script from outside of lyse, the figure doesn't get shown.

dihm: This feels really dumb, but how do you actually run analysis scripts outside of the lyse GUI? I tried running some tests that way and couldn't get past `Run` instantiation failing on the call to `self._frame_to_path(outer_frame)`. I suspect it's something simple I'm not doing right, but I couldn't find it in the docs so figured I'd ask then add to the docs for future me.

philipstarkey: @dihm it's possible it's broken. Looks like it's meant to raise a KeyError. What is it raising instead?

dihm: I'm getting this traceback running my dumb script in the same folder as the shot via `python script.py`

```shell
Traceback (most recent call last):
  File ".\LyseDecoratorEdgeCases_PR102.py", line 4, in <module>
    shot = Run('LyseDecoratorTestShot.h5')
  File "c:\users\naqsl\src\labscript-suite\lyse\lyse\__init__.py", line 313, in __init__
    outer_path = self._frame_to_path(outer_frame)
  File "c:\users\naqsl\src\labscript-suite\lyse\lyse\__init__.py", line 330, in _frame_to_path
    path = frame.f_globals['__file__']
AttributeError: 'NoneType' object has no attribute 'f_globals'
```

philipstarkey: Pretty sure that is meant to work. I think it's also meant to detect the filename too and use it for the HDF5 group.

The except clause that catches KeyError should trigger of you didn't have a file and were manually using lyse from an interactive terminal (in which case you have no filename and need to manually tell it the HDF5 group to use when saving results)

chrisjbillington: Looks like the functionality for `Run()` objects figuring out the group name when the script is run outside of lyse has been broken at some point. The code in `Run.__init__()` is popping out on the stack one level at a time, explicitly waiting to see when it gets to the `analysis_subprocess` module as an indicator that it has gone too far - but when running outside of lyse, `analysis_subprocess` is not used, so it never finds it.

This can be done way more simply. Some time ago we made the user's script more convincingly appear to be the `__main__` module, so we can find it that way, which will work when running outside of lyse as well. And no stack traversal. https://github.com/labscript-suite/lyse/pull/106 should fix.

PR #106, opened 2023-04-18
Resolve issue with Run object choosing group name based on script name

chrisjbillington:
Unsure since when this has been broken, but lyse routines used to work fine outside of lyse, as long as you specified the h5 file. However, prior to this fix, when instantiating a `Run` object whilst running a script outside of lyse, `Run.__init__()` fails to introspect the name of the main script, in order to select the default group name to which results should be saved.

The previous introspection code likely predates some changes we made such that traversing a call stack in order to find the top-level script is no longer necessary. The analysis routine appears in `sys.modules` as the `__main__` module, as does the top-level script when running an analysis routine from outside of lyse. So we can just use that. As before, if the `Run` object finds it is being instantiated in an unusual environment where there is no `__main__` module in `sys.modules` or it doesn't have a `__file__`, then we just skip setting the group name. We used to cough a warning about this, but that's no longer needed as the user will instead get an exception if they try to save a result, telling them they they need to set the group first.

### DIFF
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -18,7 +18,7 @@ from labscript_utils.dict_diff import dict_diff
 import os
 import socket
 import pickle as pickle
-import inspect
+from pathlib import Path
 import sys
 import threading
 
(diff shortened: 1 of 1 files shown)


lyse/__init__.py
```diff
@@ -244,48 +244,15 @@ def __init__(self,h5_path,no_write=False):
...
-            pass
-
-    def _frame_to_path(self, frame):
-        path = frame.f_globals['__file__']
-        return path
-
-    def _path_to_file_name(self, path):
-        file_name = os.path.basename(path).split('.py')[0]
-        return file_name
+        if not self.no_write:
```
dihm, line 247: These if statements could be combined now.

dihm (approved):
Confirmed that this fixes the issue I was seeing.

lyse/__init__.py
chrisjbillington: Excellent, -1 additional line, I love deleting code

PR #107, opened 2023-06-19
Require older tzlocal version due to breaking change in v3.0 onwards

richardt94:
When installing lyse in a fresh Python environment, pip automatically installs the latest version of `tzlocal` (5.0.1 at the moment). The call to `tzlocal` in `dataframe_utilities.py` is using the old API for the package, which was deprecated in v3.0. This PR resolves the issue by setting a restriction on the `tzlocal` version in `setup.cfg`. Alternatively the code could be updated to use the newer `tzlocal` API (and a minimum version added to `setup.cfg`).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ install_requires =
   qtutils>=2.2.2
   runmanager>=3.0.0
   scipy
-  tzlocal
+  tzlocal<3.0
   zprocess>=2.2.2
 
 [options.entry_points]


dihm:
@richardt94 Thanks for this report! I mostly work with conda installs and conda likes to slow-roll these kinds of changes (conda still only ships tzlocal=2.1 on official channels).

I'd prefer to fix this than tie things down to an old version. With some quick testing, it appears that the `pandas.Timestamp(timestr, tz=tz)` call in lyse can actually handle the raw output of `tzlocal.get_localzone()` for either the deprecated tzinfo or new zoneinfo objects (ie don't need to manually extract the zone string in any case). 

Could you change this line
https://github.com/labscript-suite/lyse/blob/c462204912cb0a1cf59f3aee6929a5385c96b72e/lyse/dataframe_utilities.py#L25
to be `tz = tzlocal.get_localzone()`?

If that works in your testing, could you update this PR and I'll go ahead a merge and make a release for our pip users out there.

dihm:
I've confirmed that the above fix works from both tzlocal 2.1 and 5.0.1

I'm going to make a separate PR to fix this.

PR #113, opened 2024-02-06
Added missing close event for pulldown menu.

ispielma:
The `exit` pulldown menu was not connected to a signal.  Fixed.


dihm:
A testament to how many people use the window X to close lyse instead of the pull-down menu.

Thanks for the fix!

PR #115, opened 2024-02-09
Found some trivial errors in the lyse doc strings while looking at startup messages

ispielma:
The follow error appears in the lyse startup messages:
```
/Users/ispielma/Python/Labscript/lyse/lyse/__init__.py:596: SyntaxWarning: invalid escape sequence '\*'
/Users/ispielma/Python/Labscript/lyse/lyse/__init__.py:758: SyntaxWarning: invalid escape sequence '\*'
/Users/ispielma/Python/Labscript/lyse/lyse/__init__.py:775: SyntaxWarning: invalid escape sequence '\*'
```
I guess at some point somebody thought that `'*'` needs an escape sequence (which it doesn't).  This pull request fixes this along with a spelling error.

### DIFF
--- a/lyse/__init__.py
+++ b/lyse/__init__.py
@@ -513,7 +513,7 @@ class Run(object):
 
     @open_file('r')
     def get_wait(self, name):
-        """Returns the wait paramteres: label, timeout, duration, and time out status.
+        """Returns the wait parameters: label, timeout, duration, and time out status.
 
         Args:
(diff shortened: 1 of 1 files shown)


dihm (approved):
These may have been put in there for the docs to render correctly, but it looks like they are building fine now with newer sphinx, so good enough for me.

Thanks!

PR #119, opened 2024-02-24
Initial refactoring

ispielma:
This is a real pull request that adds no functionality.  Instead it is a first step for any large-scale refactoring of the `lyse` code base.  All that was really done was pulling out the functional code from `__main__.py` into `main.py`.  Even this is inherently useful because other files / modules can import from 'main.py', but owing to shamefully poor language design `__main__.py` cannot be imported from.  For may larger aggregate pull request, this is why I refactored.

Doing this was slightly non trivial because `__main__.py` made liberal use of global variables that had to be transformed to method / function arguments and class variables. 

Note that this is not intended to be a complete refactor, but any next steps should be pretty easy with the global scope variables banished.

@philipstarkey Is this about the size you are looking for in a more-easy-to-audit pull request?


dihm:
Thanks for getting this broken down! This is definitely the right size and scope Before doing a full review, I have a few higher level comments.

While I'd normally say to each their own, I don't think the amount of shade thrown on python is (fully) warranted. Now I'm probably equally biased the other way here, but I think some of your judgements fly in the face of standard python conventions and will make this code harder to work on longer term.

1. Since python has made the choice that `__main__.py` is the only entry point for a stand-alone module, it actually is excellent design to prevent imports from it because it prevents all but inevitable circular imports. After all, why would something import from `__main__.py` and not ultimately need to be reimported back for actual use? Only separate API code would satisfy that assumption, but it should be in a separate file anyway.
    -  I think wholesale moving everything from `__main__.py` to `main.py` is not as much of an improvement as it could be. Really, I'd like to see the GUI code broken up into descriptive modules (ie Analysis Worker stuff in one, dataframe stuff in another, the webserver on its own, etc).
    - Surely some of these items don't actually need to be moved from `__main__.py`, right? Do you really intend to import `Lyse` or `LyseMainWindow` somewhere else? I'd say anything that is only going to be used in `__main__.py` (long term) should stay there.
2. This is more of a personal opinion, but I think "avoid globals at all costs" is scaremongering propaganda from the functional programming gang, especially in python. Because everything is an object, a "global" is just a module level attribute and in many cases is functionally equivalent to a class attribute. I personally don't feel a ton of motivation to manually track and pass around a handle to a singleton instance that is only used in a single module that is going to be modified in place via side effects anyway. 
    - Ultimately moot since splitting up functionality between files breaks an assumption making this change necessary. Mostly just wanted to caution against casually going against python conventions since they are often conventions for a valid reason. And at the end of the day, PRs are easier to get reviewed and merged safely if we can agree on conventions and instead focus attention on the hard stuff.

ispielma:
@dihm In terms of the actionable part of this review I can put the MainWindow back in `__main__.py` since it is unlikely to be imported.  I am **more than happy** to break the new `main.py` into smaller files, I didn't do that before because I was trying to keep the commit modest in size.

ispielma:
Regarding the other topics I don't want to start a philosophical war.

1. The inability to import from `__main__.py` : perhaps this is a Python documentation or error reporting issue.  The core problem is that the line `from lyse.__main__ import ...`  from within a `.py` file in the lyse module doesn't behave as expected, and no error or warning is raised.
2. My position is that could should be as easy as possible to understand by a 3rd party.  
   a. So globals such as `I_LOVE_GLOBALS = False` are great.  The common standard of all caps denotes them as some sort of global and they by convention are defined the top of a file, making them easy to find.
   b. The problem is hidden globals.  For example in `__main__.py` the global `qapplication` was defined at the bottom of the script and inside the `if __name__ == "__main__":` block.  This both makes it hard to understand the operation of the individual classes that reference qapplication in isolation without a wholistic understanding of the program and it is a bug because qapplication is not defined in the global scope `if __name__ != "__main__"`
3. Thinking of amusing bugs. Run `python` and type `import lyse.__main__` at the prompt.  You will get the lyse spash window, but nothing else (because `__name__ != "__main__"`).

ispielma:
Update: I put `Lyse` and `LyseMainWindow` back into `_main__.py` and also fixed the amusing bug (3.) above.  In doing so I realized that this combination actually defeats the splash window (because modules are loaded before the splash window opens as I had to move that to the `if __name__ == "__main__"` code block).  By having `__main__.py` nearly empty there are no required imports outside of `if __name__ == "__main__"` so we can be assured that splash does what we expect it to.

ispielma:
And split "mostly gui" code into more files.  This should pretty much complete the refactoring of what was `__main__.py`.

dihm (changes requested):
Thanks for clarifying some things for me. I hadn't looked closely enough to realize how `qapplication` was defined. I agree that isn't great. It's a good thing to fix here while we are moving things around anyway.

As for `import lyse.__main__`, while I have little doubt documentation and behavior could be improved here at a language level, I still stipulate that it should never be done. So I wouldn't really classify unusual things happening when one does it as a bug and I don't think we should structure our code to protect against it. It shouldn't be done in the first place. In fact, I see the whole purpose of this PR is to ensure that action never has to happen.

This is relevant to the only real concern I have with the PR, namely that all these imports in `__main__.py` are decoupled from their usage. It means tooling can't ensure imports are used/missing and therefore ensuring that import list is accurate is a purely manual process (lyse will still work fine even if we add extra imports or take needed ones off). I really don't like it. 

I'd much rather move current `main.py` stuff back into `__main__.py` and continue to rely on the import side effect to handle the splash. I don't see us losing any functionality, and it conforms to standard conventions so we are less likely to surprise future developers. Unless I'm missing something else, the only thing the current implementation is solving is allowing `import lyse.__main__`, but I don't think we should support that anyway.

Now, perhaps I am missing something. I am assuming that `Lyse` and `LyseMainWindow` do not need to be imported elsewhere for any reason in the future. My imagination may be limited, but I just don't see a valid reason for needing to do that that doesn't also entail a major structural change to how lyse works. And if that is the case, we should discuss it (ideally as another PR/Issue/etc, so this can move along).

Pretty sure my only other meaningful change request is some minor adjustments to the file structure. I think `analysis.py` is kind-of vague given `analsis_subprocess.py` exists. It also is only used in `widgets.py` for use with in `RoutineBox` (via an undeclared import, which is incidentally leading to a circular dependency since `analysis.py` has to import `RoutineBox`). I'd advocate for having a `routines.py` file that has `RoutineBox` and `AnalysisRoutine` in it. I'd also advocate moving more of the QT widgets from `filebox.py` into `widgets.py` (basically any class that only has QT dependencies and no specific lyse logic baked in should move to `widgets.py`).

Finally, I've noticed there are a few lingering unused imports in the files. Given that I'm asking for some changes here I won't list them all out, but an example is `lyse.analysis` and `qtutils.icons` in `filebox.py`. Annoyingly, my tooling is not catching the `lyse.analysis` unused import, but maybe that means something subtle is going on. Tread with caution I suppose.

ispielma:
I am going to decouple some of these comments.  First regarding the imports and SplashWindow.  The SplashWindow is pure eyecandy that gives the user something to see while Lyse starts.  Prior to this pull request pretty much everything that lyse was going to import was being also imported in `__main__.py`  as a result the SplashWindow could give some indication about the progress of importing all of these imports.  By pulling most/all of the code out of __main__.py there were very few imports and the SplashWindow therefore had few updates.  My solution was to recover the old behavior by manually importing everything that was imported in `__main.__.py` prior to the refactoring.  Personally I don't care about this behavior, but perhaps somebody likes it.

ispielma:
Unfortunately `import qtutils.icons` needs to be present as it performs some sort of magic that allows the .ui files to reference that fugue icon set without defining the exact path in QtDesigner.  I would like to add an explicit option to the ui loader such as
```
self.ui = loader.load(os.path.join(LYSE_DIR, 'user_interface/main.ui'), LyseMainWindow(self), fugue_icons=True)
```
to make this behavior explicit and avoid the magical import.

ispielma:
The last set of changes should fully resolve @dihm's points.

dihm (changes requested):
@ispielma This is really good. Have two very small comments then I suspect it is ready to go. 

I would like to stress test it a little more than the dummy shots on my home rig, but I won't have time until later this week. Hopefully that will give @philipstarkey sufficient time to look things over if he wants to weigh in. Otherwise, I think we'll be good to merge by Friday.

lyse/routines.py
```diff
@@ -0,0 +1,557 @@
...
+# Labscript imports
+from labscript_utils.qtwidgets.headerview_with_widgets import HorizontalHeaderViewWithWidgets
+
+# qt imports
+from qtutils.qt import QtCore, QtGui, QtWidgets
+from qtutils import inmain_decorator, UiLoader, DisconnectContextManager
+
+from lyse import LYSE_DIR
+import lyse.widgets
+import lyse.routines
```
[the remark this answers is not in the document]
ispielma: Removed.  It is interesting that the code worked with that there since it is self referential.

lyse/communication.py
```diff
@@ -0,0 +1,125 @@
...
+
+# Labscript imports
+from labscript_utils.ls_zprocess import ZMQServer
+import labscript_utils.shared_drive as shared_drive
+
+# qt imports
+from qtutils import inmain_decorator
+
+# Lyse imports
+from lyse import _rangeindex_to_multiindex
```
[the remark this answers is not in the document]
ispielma: Since this is generally a refactoring point there is not reason not to do this now, so I did and also renamed it to be just `rangeindex_to_multiindex` since there is no reason to have an underscore if it is stored in `dataframe_utilities`.

lyse/routines.py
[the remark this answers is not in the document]
ispielma: Since it looks like this package of changes is pretty much tied up, I will move on to worker sub-processes and windows.  In the short term I will introduce a *new style* worker that will be dispatched if the worker script starts with a comment something like 
```python
# analysis_version = "2.0"   
```
this should allow easy backward compatibility.

lyse/communication.py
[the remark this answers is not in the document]
dihm: @ispielma This change broke the package on import. Needs to be either `from .dataframe_utilities import ...` or matching the style we have been using `from lyse.dataframe_utilities import ...`

dihm:
I also forgot, there is probably a little work that needs to be done on the documentation build to track the new changes. If you wanted to sort those out, I would appreciate it. If not, I'll get them sorted this week and make a PR to your branch with the changes.

ispielma:
I am not that familiar with the documentation system (in terms of what is automatic and what is manual), but I will have a look.

lyse/dataframe_utilities.py
```diff
@@ -18,6 +18,31 @@
...
+def rangeindex_to_multiindex(df, inplace):
+    if isinstance(df.index, pandas.MultiIndex):
+        # The dataframe has already been converted.
+        return df
+    try:
+        padding = ('',)*(df.columns.nlevels - 1)
+        try:
+            integer_indexing = _labconfig.getboolean('lyse', 'integer_indexing')
+        except (LabConfig.NoOptionError, LabConfig.NoSectionError):
+            integer_indexing = False
```
dihm, line 30: Just had a go at testing this in lab and had a failure here. Neither `_labconfig` nor `LabConfig` are defined in `dataframe_utilities.py`

dihm (changes requested):
Found another small issue testing the refactored code in the lab.

dihm:
> I am not that familiar with the documentation system (in terms of what is automatic and what is manual), but I will have a look.... OK so I have it generating the API in a way that is no worse than before, meaning that many functions / classes are not documented, but I don't think that will be its own pull request.

Thanks for sorting that out. Docs are definitely spotty, but at some level most people don't need to see the GUI docstrings anyway. In fact, we should probably consider splitting the `lyse` autosummary from all the others (ie divide up API and GUI docstrings) just to make it clearer what is actually important.

lyse/dataframe_utilities.py
ispielma: Thanks for finding this.  I don't understand what is going on here that these errors are not appearing when I am doing local testing (nor why my editor is not seeing them).

dihm: This is still only partially fixed. `_labconfig` is not defined in `dataframe_utilities.py` either. 

Not sure why things aren't tripping for you. This particular error pops for me when running a multi-shot analysis after importing a large batch of runs (which had multiple single-shot analyses run on them). I suppose there is also a chance that we have slightly different labconfig values for lyse that are leading to slightly different behavior.

ispielma: This fix was a bit less trivial (see commit log).  To avoid circular imports I moved `_labconfig` and `_lyse_port` to utils.py and changed them to `LABCONFIG` and `LYSE_PORT` to mark them as globals.  For the same reason I moved `LYSE_PATH` to the same place.  For testing I ran on both a PC and a MAC and did both single and multishot scripts, so hopefully there are no more problems.

ispielma:
In working with `__init__.py` I realized that its exports are somewhat uncontroled.  Meaning that everything in the namespace will be imported with `import lyse` or `from lyse import *`.  To be clear, an example of this is
```python
from lyse.dataframe_utilities import get_series_from_shot as _get_singleshot
from labscript_utils.dict_diff import dict_diff
```
in this case when one does `import lyse` the function `dict_diff` will be entered into the name space and `lyse.dict_diff` will be defined.  Clearly the authors knew about this which is why we see `... as _get_singleshot`.

I _strongly_ suggest that I amend this pull request to also create a file `lyse_api.py` and move almost the whole content of `__init__.py` there and have `__init__.py` instead be more like
```python
from lyse.lyse_api import ...
__all__ = [...]
```
this will make explicit what is being exported and provide better control over the namespace, and the use of `__all__=[...]` would define what we want from a `*` import.

lyse/analysis_subprocess.py
```diff
@@ -432,17 +445,12 @@ def reset_figs(self):
...
-    import lyse
-    from lyse import LYSE_DIR
     lyse.spinning_top = True
-    import lyse.figure_manager
+
     lyse.figure_manager.install()
 
-    from matplotlib.backends.backend_qt5agg import NavigationToolbar2QT as NavigationToolbar
+    # Where are pylab features used?  Should this be here?
     import pylab
```
dihm, line 453: I'm not entirely sure why this is here. My best guess is that importing pylab has side-effects that alter the default plot styling. If so, that is something that we could fix much better using a mpl style file. 

Ultimately, I'm on board with getting rid of all `pylab` imports across labscript as it is basically an abandoned interface to matplotlib. But those changes are often technically breaking since users might be using the import incidentally (we do recommend them in the docs in a number of places). Which is to say this should be the subject of another PR.

dihm (commented):
As for moving the API to another file, I'm not totally sold on the benefit. We should definitely define a `__all__`, but we can just do that in place. The incidental imports still show up under `import lyse` (like sys, os, etc), but I'm less concerned about that in that situation.

In any case, given the docs have recommended `from lyse import *` for forever, this would be a breaking change, so we should save it for another PR.

lyse/analysis_subprocess.py
```diff
@@ -12,6 +12,8 @@
 #####################################################################
  
 import labscript_utils.excepthook # I do magic stuff, so import must be in place
+import labscript_utils.h5_lock, h5py
+
 from labscript_utils.ls_zprocess import ProcessTree
```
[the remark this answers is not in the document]
ispielma: I can do that now, but I think in the next pull request I will end up undoing it.  The reason is that this code 
```
    process_tree = ProcessTree.connect_to_parent()
    to_parent = process_tree.to_parent
    from_parent = process_tree.from_parent
    kill_lock = process_tree.kill_lock
    filepath = from_parent.get()
```
in `if __name__ == '__main__':` leaves `kill_lock` as a global that is then used in a class defined outside of this of statement.  In my next group of pull requests I will be moving the definition of `process_tree` into `AnalysisWorker` which really should own everything about it.

lyse/analysis_subprocess.py
ispielma: Given this I will not make any of my suggested changes to __init__.py at this time.

lyse/analysis_subprocess.py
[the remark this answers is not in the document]
dihm: Fair enough. I say leave it as is then.

dihm (dismissed):
I'm happy with where this. I'll give @philipstarkey a little more time to look it over before merging, now that the bugs have been sorted out.

lyse/__main__.py
```diff
@@ -2328,19 +509,15 @@ def delete_items(self, confirm):
...
     qapplication = QtWidgets.QApplication.instance()
     if qapplication is None:
         qapplication = QtWidgets.QApplication(sys.argv)
     qapplication.setAttribute(QtCore.Qt.AA_DontShowIconsInMenus, False)
-    app = Lyse()
 
-    # Start the web server:
-    splash.update_text('starting analysis server')
-    server = WebServer(app.port)
-    splash.update_text('done')
```
philipstarkey, line 2343: The relocation of the `WebServer` into the `Lyse` class, and instantiating it before the GUI is set up, means that any messages received immediately on start could crash lyse when the `WebServer.handler` method tries to access something about the UI that isn't instantiated yet.

The previous implementation worked such that messages could be received, and placed in appropriate event queues before the Qt loop was even started.

I suspect, given the auto-retry behaviour of BLACS, that there would be scenarios where the crash will occur (it's a bit of a race condition though so may not be easily replicable)

lyse/communication.py
```diff
@@ -0,0 +1,125 @@
+#####################################################################
+#                                                                   #
+# /main.py                                                          #
```
philipstarkey, line 3: Incorrect filename (there could be other instances of this - please check all the new files)

lyse/utils.py
```diff
@@ -0,0 +1,121 @@
...
+
+LYSE_DIR = os.path.dirname(os.path.realpath(__file__))
+
+# get port that lyse is using for communication
+try:
+    LABCONFIG = LabConfig(required_params={"ports": ["lyse"]})
+    LYSE_PORT = int(LABCONFIG.get('ports', 'lyse'))
+except Exception:
+    LABCONFIG = None
+    LYSE_PORT = 42519
```
philipstarkey, line 32: I'm not super comfortable with this. Instantiating a `LabConfig` just because you import `lyse.utils` has a bad code smell to it. It also isn't used by anything in this file. Seems like it was just move to fix some sort of circular dependency issue?

I'm sure there are a few possible solutions. One that may be the simplest is just to move the import of the labconfig inside the function that uses it, so that it can import it from `lyse`. That's perfectly acceptable to do, and a good way to break circular dependencies if one part only needs something at run time, not load time.

philipstarkey (changes requested):
I've just reviewed the changes for now. I'm also hoping to find time to check out the new code and give it a look over to see if there is anything that stands out about the structure (which can be hard to see with all of the noise from the diff)

dihm:
@ispielma I believe I have finally cleared off some time from my calendar, and this PR is well beyond over due to be finished and merged. If at all possible, I'd like to get it done in the next week or two. If that timing doesn't work for you, I can take over (by pushing to your branch or creating a new PR based on this one) and try to address Phil's review so this can finally be done. Let me know what your preferences are.

dihm:
@ispielma I have started working on some commits to address Phil's review. Are you OK with me pushing to your branch directly? I think it would be best to keep the work in a single PR (if possible, github rules for pushing to forks can be odd sometimes).

lyse/__main__.py
```diff
@@ -1,272 +1,95 @@
...
 import queue
 import warnings
+import signal
 
 # 3rd party imports:
 splash.update_text('importing numpy')
 import numpy as np
-splash.update_text('importing h5_lock and h5py')
-import labscript_utils.h5_lock
-import h5py
```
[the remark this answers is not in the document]
dihm: I believe the answer is the main GUI does not perform any direct accesses to h5 files. These accesses are provided by the main API, filebox, and dataframe utilities. The import is therefore handled within the `core Lyse modules` portion.

I did, however, find an upatched h5py import in filebox that will be fixed.

lyse/communication.py
dihm: I have checked and fixed all of these.

lyse/__main__.py
```diff
@@ -1,272 +1,95 @@
...
+# Simplified BSD License. See the license.txt file in the root of   #
+# the project for the full license.                                 #
+#                                                                   #
+#####################################################################
+
+"""
+Top level Lyse GUI
 """
 import os
-import labscript_utils.excepthook
```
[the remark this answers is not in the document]
dihm: I'll relocate this one up.

lyse/__init__.py
```diff
@@ -10,7 +10,8 @@
 # the project for the full license.                                 #
 #                                                                   #
 #####################################################################
-"""Lyse analysis API
+"""
+Lyse analysis API
```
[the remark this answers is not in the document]
dihm: I've made these consistent, and added to files that didn't already have one.

lyse/__init__.py
```diff
@@ -10,7 +10,8 @@
...
 """
 
 from lyse.dataframe_utilities import get_series_from_shot as _get_singleshot
```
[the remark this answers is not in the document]
dihm: I've fixed this one.

lyse/__init__.py
```diff
@@ -36,7 +37,13 @@
 from labscript_utils.ls_zprocess import zmq_get
 
 from labscript_utils.properties import get_attributes, get_attribute, set_attributes
-LYSE_DIR = os.path.dirname(os.path.realpath(__file__))
+
+# lyse imports
+import lyse.dataframe_utilities
+import lyse.utils
```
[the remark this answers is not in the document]
dihm: I've had a go at implementing this and did end up settling on your suggested divisions. All said and done, I think there is a decent argument for moving `dataframe_utilities` into the submodule _or_ flattening the structure again and having `utils`, `gui_utils` and `worker_utils` at the top level. I am leaning towards the latter, but I think there is a more fundamental issue to fix first (I'll comment the code elsewhere) before sorting out stylistic preferences like this one.

lyse/analysis_subprocess.py
```diff
@@ -433,17 +445,12 @@ def reset_figs(self):
...
     os.environ['MPLBACKEND'] = "qt5agg"
 
-    import lyse
-    from lyse import LYSE_DIR
     lyse.spinning_top = True
-    import lyse.figure_manager
+
     lyse.figure_manager.install()
 
-    from matplotlib.backends.backend_qt5agg import NavigationToolbar2QT as NavigationToolbar
```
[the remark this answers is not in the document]
dihm: I actually think this is OK. The comment specifically calls out that the install needs to occur before importing `matplotlib.pyplot`. I don't think there are issues with importing parallel submodules from elsewhere in `matplotlib`.

lyse/__main__.py
dihm: I believe I understand the issue here. Would moving the start of the `WebServer` to the end of the GUI init be sufficient? I believe that would recover the older behavior. I don't think we can go back to the old behavior exactly since it relied on globals-like behavior which is a bit difficult to reason about.

lyse/__init__.py
```diff
@@ -55,13 +61,6 @@
 # a flag to determine whether we should wait for the delay event
 _delay_flag = False
```
dihm, line 62: These variables are causing quite a bit of grief for me. We have a decent number of places that need them across the modules, and I think it is generally considered a bad code smell to import from the module `__init__`. Places these global-like values are used:
- `analysis_subprocess`
  - `_delay_flag`
  - `delay_event`
  -  `spinning_top`: redefinition
  - `_updated_data`
  - `path`: redefinition
  - `plots`: redefinition
  - `Plot`: redefinition
- `figure_manager`
  - `spinning_top`
- `utils.worker`: I've moved stuff here from `__init__` on my working branch; I am slowly realizing that I don't really understand their purpose
  - `spinning_top`: `register_plot_class`, `get_plot_class`, and `delay_results_return`
  - `_plot_classes`: `register_plot_class` and `get_plot_class`
  - `_delay_flag`: `delay_results_return`
- `__init__`
  - `path`
  - `spinning_top` in `Run.save_result`
  - `_updated_data` in `Run.save_result`

I'm still trying to grok exactly how they are used and what their purpose is, as it isn't obvious to me why they are defined in this way. Most of them only appear to be necessary if you are using a script within the lyse GUI, and they wouldn't even be called with the plain API. I suspect I am unaware of some standard workflow. Do people commonly run their lyse scripts in the GUI, then also run them as stand-alone batch scripts unmodified (hence why `register_plot_class`, `get_plot_class` and `delay_results_return` are defined in the API)?

Right now, I am inclined to move these global-like variables down to `utils.worker`. Ideally I'd also like to consider other ways for the GUI to communicate to the analysis subprocesses that doesn't rely on passing through module variables. I'm probably not the man for that job, but I feel like there must be a better way. Perhaps manual injection into the `AnalysisWorker.routine_module.__dict__`?

dihm: Spending a little more time with this, I don't see the purpose of `lyse.Plot` or `lyse.plots`. They do not appear to be accessed anywhere, and they are only set to something meaningful within `AnalysisWorker.do_analysis`. I am fairly confident they can just be removed. Am I missing something?

ispielma: Those are part of the next step of my Lyse changes that seem to have leaked into this refactor.  Their function is associated with making the lyse plot windows "first class" windows whose positions are when lyse shuts down.  This is the highest priority Lyse request in my group.

dihm: OK, though they appear to be on mainline already. Doesn't really matter where they came from I suppose, I'll plan on keeping them in. 

Would it be possible to implement figure position tracking independent of lyse globals? Chunking out distinct functionality, where feasible, goes a long way in ensuring these reviews don't drag on into absurdity.

dihm:
OK, I've gone ahead and pushed minor edits directly to the refactor branch. I've also created a PR to Ian's branch (ispielma/lyse#1) that contains more disruptive changes. Hopefully it makes it easier to see/discuss what they are relative to these changes as they presently exist.

lyse/__init__.py
ispielma: Since I have not visited this in about a year, I will use my old code as reference, but rewrite the implementation based on the current state of Lyse.

ispielma:
Sorry for the delay.  I went over this and found only one issue which was a leftover reference to `lyse.spinning_top`.  Other than that I am happy with the changes.  I ran some shots with it and they were all OK, but I have not deployed in a running lab.

In addition, I moved all of these changes into my feature development branch (and simplified that branch).  I ran with that branch for a while in the lab, and I am going to switch a whole team to use it to get some more usability feedback before reviving that pull request.

lyse/__init__.py
```diff
@@ -43,29 +38,18 @@
...
-
+from lyse.utils.worker import spinning_top, _updated_data, register_plot_class, delay_results_return
+
+# note: path is injected into the script namespace by AnalysisWorker at runtime
+def __getattr__(name):
+    if name == 'path':
+        from lyse.utils.worker import path
+        warnings.warn("'path' is now automatically injected into the script namespace. "
+                      "Importing it from 'lyse.path' will be deprecated.",
+                      warnings.FutureWarning)
```
dihm, line 49: This warning definition is wrong. It should just be `FutureWarning`

lyse/__init__.py
```diff
@@ -43,29 +38,18 @@
...
-plots = {}
-# A threading.Event to delay the 
-delay_event = threading.Event()
-# a flag to determine whether we should wait for the delay event
-_delay_flag = False
-
-if len(sys.argv) > 1:
-    path = sys.argv[1]
-else:
-    path = None
```
dihm, line 68: Looking back over this, I have a concern that we are removing a workflow (ie running `python analysis.py h5_path.h5` to run single shot analyses without needing to run lyse. Personally I don't think it is much of a loss, but I also don't want to break too much if people actually use that workflow.

Supporting it will be slightly annoying, but certainly possible. @ispielma Do you have any input?

lyse/__init__.py
dihm: I would just fix it, but it appears I can't push commits to your branch anymore (probably for good reason).

dihm (changes requested):
One minor fix remaining, and a potential remaining backwards compatibility decision to make. 

Once sorted, I'm going to finally merge this long over-due upgrade. I'll plan on sitting on it for at least a week before tagging a release so I can test in lab here a bit more and sync up with some lingering fixes to other components.

lyse/__init__.py
ispielma: I don't think we should support this.  My argument is that the whole way lyse works to modify the namespace is a really bad idea (TM) anyway.  One of the back-end things I am working on is adding a second lyse workflow that removes all the magic that is currently under the hood (with the plan to do the same to the format of labscript scripts.

So I'd rather spend time on this rather than yet more magic under the hood to support a potential use case.

dihm: I am inclined to agree. The inter-process communication here isn't my favorite. It is probably sporting to add a helpful hint in case someone actually does this so they know who to blame for borking their workflow.

Though the more I think about the changes I've made, I'd like to modify how `path` is bodged. One of my personal goals is to remove all star import from labscript, including in examples and docs. `from lyse import *` is a common paradigm, and the code I wrote currently would have things like `run = lyse.Run(path)`, which mixes the magical syntax. I think I'd just as soon keep the underlying machinery but re-expose `path` as `lyse.path` so people at least are clear on where the magic is coming from.

I'll make another PR against your branch.

dihm: @ispielma I've made a few final tweaks that ensure existing star imports do not break. While I really want to nuke star imports, breaking them without warning doesn't feel right. 

Anyway, we've been testing these changes for a bit without issues, so I'm happy to merge. If you can merge my PR to yours, that would be great. Otherwise, I'll go ahead and merge as-is and move my patches over directly.

PR #135, opened 2026-03-20
Accept kwargs in overridden methods of Qt subclasses

chrisjbillington:
This is necessary as some PyQt6 machinery will call these methods with keyword arguments


chrisjbillington:
Closing this as I'll commit it directly to #134 instead

